PR #15, opened 2023-04-21
macOS getting started added

PeterTurcan:


alandefreitas:
At some point, we need to separate the official instructions for installing boost from the instructions using a package manager (vcpkg, brew, apt-get, ...). While both approaches are valid and most users will use a package manager, the official procedure is the only procedure we support, the only one over which we have control to fix bugs, the only one that will provide users with the latest stable version, and the one package manager recipe writers need to create the packages, to begin with.

The getting started instructions also have some redundant content in that, for instance, [Getting Started with Linux](https://15.site-docs.prtest2.cppalliance.org/user-guide/getting-started-with-linux.html) talks about package managers and [Use Boost with Linux and a Package Manager](https://15.site-docs.prtest2.cppalliance.org/user-guide/use-boost-with-linux-package-manager.html) include the instructions again. In some cases, like [Getting Started with Linux](https://15.site-docs.prtest2.cppalliance.org/user-guide/getting-started-with-linux.html) and [Use Boost with Linux and a Package Manager](https://15.site-docs.prtest2.cppalliance.org/user-guide/use-boost-with-linux-package-manager.html), the first included instructions for a package manager (`sudo apt install libboost-all-dev`) while the later is using vcpkg but no apt-get, which is the official package manager on linux.

I don't know the best way to organize these sections hierarchically. A starting point could be "Getting started with (platform)" including only official instructions and then "Getting started with (package manager)" for each package manager. For instructions about package managers, we should include a disclaimer that we don't maintain or support these.

When I talk about the official instructions, I refer to both the bootstrap.bat/sh method described in the current documentation and the [CMake](https://github.com/boostorg/cmake#building-boost-with-cmake) method Peter developed but never made it to the getting started documentation.

alandefreitas (approved):

vinniefalco:
@alandefreitas approving the PRs is not enough you have to actually merge them (fast-forwards only please)

alandefreitas:
Oh... I thought it was usual to have some discussion, etc... OK.

PR #38, opened 2023-05-02
[WIP] Add Antora Lunr Extension

ashtum:
This PR adds [lunr-extension](https://gitlab.com/antora/antora-lunr-extension) to `libs.playbook.yml` and `site.playbook.yml`.
Minor changes have been applied to antora-ui to add a checkbox for limiting search scopes.

The size of generated `search-index.js` for `site.playbook.yml` is 871KB.
The size of generated `search-index.js` for `libs.playbook.yml` (which currently includes mp11, predef, qvm, and unordered) is 2.1MB.
It seems impractical to use in-browser search for searching all libraries in the Boost (we should expect a search-index close to 100MB).


antora-ui/src/partials/header-content.hbs
```diff
@@ -38,10 +41,11 @@
...
-            <a class="button is-primary" target="blank_" href="https://github.com/cppalliance/site-docs/issues/new/choose">Report an Issue</a>
+            <a class="button is-primary" target="blank_"
+              href="https://github.com/cppalliance/site-docs/issues/new/choose">Report an Issue</a>
           </span>
         </div>
       </div>
     </div>
   </nav>
-</header>
+</header>
```
vinniefalco, line 51: missing a newline

alandefreitas (approved):
Looks good to me

alandefreitas:
It should be squashed though. It doesn't look like the first commit justifies itself.

libs.playbook.yml
```diff
@@ -39,6 +39,7 @@ site:
 antora:
   extensions:
     - require: ./boost.js
+    - require: '@antora/lunr-extension'
```
alandefreitas, line 42: I think just 

```yml
    - ./boost.js
    - '@antora/lunr-extension'
```

(no require)

should be fine.

ashtum: Yes as long as we have no configuration keys to pass https://gitlab.com/antora/antora-lunr-extension#user-content-index-only-the-latest-version
using map syntax will keep the consistency.

PR #40, opened 2023-05-02
Round 1 of issues with User Guide

PeterTurcan:


vinniefalco:
The links refer to 1.82.0 which is not right they need to be evergreen

PR #46, opened 2023-05-03
footer includes commit hash

alandefreitas:


vinniefalco:
And what about lib-doc? If its a tagged release it needs the commit as well

sdarwin:
Usually the short sha is the first 7 (or 8) characters of the commit sha. Not the end characters of the commit sha

PR #49, opened 2023-05-05
history section added

PeterTurcan:

PR #50, opened 2023-05-06
Boost extension

alandefreitas:
This PR adds a new `boost` inline macro to create links to a corresponding Boost C++ library. It also includes utility functions to convert strings from and to PascalCase and snake_case. These functions ensure the links are always correct in the `https://www.boost.org/libs/my_lib` pattern for the URL while the title is always in the canonical `Boost.MyLib` format used in boost. It also supports target attributes, which replace `Boost.MyLib` with whatever text is provided as a positional attribute, making it compatible with asciidoc macros that generate regular URL links.

The PR also includes a fix to the scripts that were preventing previews to be generated correctly. For instance, #49 has new content that is not reflected in its preview https://49.site-docs.prtest2.cppalliance.org/index.html because of a previous problem in the interaction between the scripts and the boost extension, that would make antora fetch and render the contents of the `develop` branch regardless of the contents of the current branch even when developing locally and using the shell scripts. This is why #49 renders the develop branch instead of the content in Peter's PR. The reason this bug might have been unnoticed for such a long time before causing us problems is probably that many people have been using their `develop` branch for almost everything at these early stages of development.

The commits in this PR are on top of Peter's #49 since it includes content we can use to test the scripts and visualize the result in the preview for this PR. In other words, there should be a [History of Boost](https://50.site-docs.prtest2.cppalliance.org/user-guide/boost-history.html) link in the sidebar, which indicates the fix worked. Peter's commit can be dropped from this PR before merging it to make the PRs independent.


anarthal:
Confirming that this solves #55.

PR #54, opened 2023-05-08
Setup Algolia DocSearch

ashtum:
#### Attributes for faceting: [Learn more](https://www.algolia.com/doc/api-reference/api-parameters/facets/)
We need to use version-specific filtering for referencing documentation, as separate documentation will be available for each Boost release.
```
lang, type, version
```

#### List of separators characters to index: [Learn more](https://www.algolia.com/doc/api-reference/api-parameters/separatorsToIndex/)   
By default, this list is empty, and the search engine ignores them. For example, it treats `C++` as `C`.
```
!#()[]{}*+-_一,:;<>?@/\^|%&~£¥$§€`'‘’“”†‡.
```

#### Crawler configuration in www.crawler.algolia.com:
`appId` and `apiKey` are removed
```JavaScript
new Crawler({
  rateLimit: 8,
  maxDepth: 10,
  maxUrls: 5000,
  startUrls: ["https://docs.cppalliance.org/user-guide/index.html"],
  renderJavaScript: false,
  sitemaps: ["https://docs.cppalliance.org/sitemap.xml"],
  ignoreCanonicalTo: true,
  discoveryPatterns: ["https://docs.cppalliance.org/**"],
  schedule: "at 14:00 on Friday",
  actions: [
    {
      indexName: "cppalliancedocs",
      pathsToMatch: ["https://docs.cppalliance.org/**"],
      recordExtractor: ({ helpers }) => {
        return helpers.docsearch({
          recordProps: {
            lvl1: [".doc > h1.page"],
            content: [".doc p, .doc td.content, .doc th.tableblock"],
            lvl0: {
              selectors: ".nav-panel-explore .context .title",
              defaultValue: "Home",
            },
            lvl2: [".doc .sect1 > h2:first-child, .doc > h1.sect0"],
            lvl3: [".doc .sect2 > h3:first-child"],
            lvl4: [".doc .sect3 > h4:first-child"],
            version: {
              selectors: ".nav-panel-explore .context .version",
            },
          },
          aggregateContent: true,
          recordVersion: "v3",
        });
      },
    },
  ],
  safetyChecks: { beforeIndexPublishing: { maxLostRecordsPercentage: 30 } },
  initialIndexSettings: {
    cppalliancedocs: {
      attributesForFaceting: ["type", "version"],
      attributesToRetrieve: [
        "anchor",
        "content",
        "hierarchy",
        "url",
        "component",
      ],
      attributesToHighlight: ["hierarchy", "content"],
      attributesToSnippet: ["content:10"],
      camelCaseAttributes: ["hierarchy", "content"],
      searchableAttributes: [
        "unordered(hierarchy.lvl0)",
        "unordered(hierarchy.lvl1)",
        "unordered(hierarchy.lvl2)",
        "unordered(hierarchy.lvl3)",
        "unordered(hierarchy.lvl4)",
        "unordered(hierarchy.lvl5)",
        "unordered(hierarchy.lvl6)",
        "content",
      ],
      distinct: true,
      attributeForDistinct: "url",
      customRanking: [
        "desc(weight.page_rank)",
        "desc(version)",
        "desc(weight.level)",
        "asc(weight.position)",
      ],
      ranking: [
        "words",
        "filters",
        "typo",
        "attribute",
        "proximity",
        "exact",
        "custom",
      ],
      highlightPreTag: '<span class="algolia-docsearch-suggestion--highlight">',
      highlightPostTag: "</span>",
      minWordSizefor1Typo: 3,
      minWordSizefor2Typos: 7,
      allowTyposOnNumericTokens: false,
      minProximity: 1,
      ignorePlurals: true,
      advancedSyntax: true,
      attributeCriteriaComputedByMinProximity: true,
      removeWordsIfNoResults: "allOptional",
    },
  },
  appId: "",
  apiKey: "",
});
```

### DIFF
--- a/antora-ui/src/css/algolia-extra.css
+++ b/antora-ui/src/css/algolia-extra.css
@@ -1,0 +1,67 @@
+.algolia-autocomplete .ds-dropdown-menu [class^=ds-dataset-] {
+  overscroll-behavior: none;
+  max-height: calc(100vh - 3.25rem);
+}
+
+.algolia-autocomplete .algolia-docsearch-suggestion--title {
+  font-weight: 500;
+}
+
+.algolia-autocomplete .algolia-docsearch-suggestion .algolia-docsearch-suggestion--subcategory-column {
+  font-weight: normal;
+}
+
+@media screen and (max-width: 768px) {
+  .algolia-autocomplete .ds-dropdown-menu {
+    min-width: calc(100vw - 2.75rem) !important;
+  }
+}
+
+.doc .image.conum > img {
+  height: 0.9em;
+  width: auto;
+  vertical-align: -0.1em;
+}
+
+#search-input-algolia-algolia {
+  color: #333;
+  font-family: inherit;
+  font-size: 0.95rem;
+  width: 150px;
+  border: 1px solid #dbdbdb;
+  border-radius: 0.1em;
+  line-height: 1.5;
+  padding: 0.25em;
+}
+
+#search-input-algolia {
+  color: #333;
+  font-family: inherit;
+  font-size: 0.95rem;
+  width: 150px;
+  border: 1px solid #dbdbdb;
+  border-radius: 0.1em;
+  line-height: 1.5;
+  padding: 0 0.25em;
+}
+
+#search-input-algolia:disabled {
+  background-color: #dbdbdb;
+  /* disable cursor */
+  cursor: not-allowed;
+  pointer-events: all !important;
+}
+
+#search-input-algolia:disabled::placeholder {
+  color: #4c4c4c;
+}
+
+#search-input-algolia:focus {
+  outline: none;
+}
+
+@media screen and (min-width: 769px) {
+  #search-input-algolia {
+    width: 200px;
+  }
+}
--- a/antora-ui/src/css/site.css
+++ b/antora-ui/src/css/site.css
@@ -18,3 +18,4 @@
 @import "tailwindcss/base";
 @import "tailwindcss/components";
 @import "tailwindcss/utilities";
+@import "algolia-extra.css";
--- a/antora-ui/src/partials/footer-scripts.hbs
+++ b/antora-ui/src/partials/footer-scripts.hbs
@@ -1,7 +1,53 @@
+{{#if env.ALGOLIA_API_KEY}}
+<script src="https://cdn.jsdelivr.net/npm/docsearch.js@2.5.2/dist/cdn/docsearch.min.js"></script>
+<script>
+  function setupAlgoliaDocsearch(){
+    document.getElementById('search-input').hidden = true;
+    document.getElementById('search-input-algolia').hidden = false;
+    var search = docsearch({
+      {{#with env.ALGOLIA_APP_ID}}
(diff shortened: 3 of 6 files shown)


alandefreitas:
I don't have a good idea of how algolia works so I don't have an informed opinion on the configuration. 
Maybe you could include some comments about what the less obvious changes in the PR do.

One problem I noticed is it only matches "C" if you search for "C++".
I also had this problem when I tested the other search box locally but it seems like it didn't affect site-docs.
It affects it now with algolia although the my local experiment suggests this is not only about angolia.

The design of the algolia search bar looks much uglier than the other bar. At least to my bad taste, because I'm not a designer at all. The design is too busy and draws too much attention from the content to the search bar.
My bad design instinct is this could be something like the old regular search bar with the text "Search (ctrl + k)" in the placeholder attribute.

alandefreitas:
Also about the matches:

image

"User Guidedefault" is definitely not the name we want here. Either "User Guide" if we're just searching this branch or "User Guide (master)", "User Guide (branch-name)" or whatever, if we're searching all branches and the current branch is not "branch-name".

ashtum:
@alandefreitas 
> One problem I noticed is it only matches "C" if you search for "C++".
Yes, this is a common problem with search engines, I'll investigate to see if there is any remedy for it.

> The design of the Algolia search bar looks much uglier than the other bar.
I agree, this is the default style coming with the DocSearch, we can change it later on if we decide to go forward with it. 

> "User Guidedefault" is definitely not the name we want here.
That was an issue in the crawler configuration. Now it is working correctly.

alandefreitas:
> Yes, this is a common problem with search engines, I'll investigate to see if there is any remedy for it.

Yes. I still don't understand why the old solution worked. Both solutions failed to search for C++ for me locally. 

> we can change it later on if we decide to go forward with it.

I guess how soon you would do it depends on how complex the task is, which only you can evaluate. But I don't see any reason why we wouldn't go forward with it. Let me know if there's something I'm missing here.

> 
@alandefreitas

One problem I noticed is it only matches "C" if you search for "C++".

Yes, this is a common problem with search engines, I'll investigate to see if there is any remedy for it.

The design of the Algolia search bar looks much uglier than the other bar.

I agree, this is the default style coming with the DocSearch, we can change it later on if we decide to go forward with it.

> That was an issue in the crawler configuration. Now it is working correctly.

Nice.

Please note you'll need to rebase this PR. 
Please ignore my comment if this sounds too trivial. 
I don't know how familiar with git everyone is so I initially assume the worst and update my expectations accordingly.

ashtum:
> Yes. I still don't understand why the old solution worked. Both solutions failed to search for C++ for me locally.

It is fixed now, there is a `separatorsToIndex` parameter in the configuration which includes special characters in the search.

> I guess how soon you would do it depends on how complex the task is, which only you can evaluate. But I don't see any reason why we wouldn't go forward with it. Let me know if there's something I'm missing here.

I meant that we might ultimately decide not to use Algolia. Therefore, investing time in the styles at the beginning might not be worthwhile.

ashtum:
@alandefreitas This is ready to merge.

PR #58, opened 2023-05-09
getting started guide uses tabs

alandefreitas:
This PR modifies the getting started guides to use tabs. The content is structured in a single flow where only commands that differ between the platforms are separated into tabs. The tabs are synced so that clicking on a tab that says "Windows" will show instructions for Windows on all other tabs on the page. Content has been merged i an way that's easier to follow but it still needs as much review as before.


vinniefalco:
image

user-guide/modules/ROOT/pages/getting-started.adoc
```diff
@@ -0,0 +1,886 @@
...
+
+NOTE:: We break the build and install into separate steps, in order to get better diagnostics when building the libraries, and to suppress too many messages when installing the libraries.
+The single command `.\b2.exe --prefix=<YOUR-INSTALL-FOLDER> -d0 install` will both build and install, but suppresses too many messages during the build process.
+If something goes wrong, it will probably be during the build process, and we want to know about it.
+
+The build process creates all variants of each library.
+For example, there may be _Release_ and _Debug_ variants, or possibly _Multithreaded_ and _Static_ runtime variants.
+Also, libraries that are _header-only_ (in other words, do not require being built) are copied to your installation folder.
+
+TIP: Consider using the `--show-libraries` and `--with-libraries=library-name-list` options if you want to limit the wait instead of building everything.
```
anarthal, line 248: AFAIK `--with-libraries=x,y` doesn't work. It needs to be `--with-libx --with-liby`.

user-guide/modules/ROOT/pages/getting-started.adoc
```diff
@@ -0,0 +1,886 @@
...
+--
+
+CMake::
++
+--
+You can also build the Boost libraries with CMake.
```
anarthal, line 254: If I'm not mistaken, if you build Boost with CMake, you end up getting not just a different layout, but a different set of CMake targets. So if you built Boost with b2 and you're using Boost.Asio, you will link against `Boost::headers`, while if you built it with CMake, you will link against `Boost::asio`. I see 1.82 version of cmake will also provide `Boost::headers` for compatibility.

I'm not sure whether we're really prepared to include cmake at this point - wouldn't just be cleaner to stick to the official method for now (you can then use Windows/Unix tabs), then attempt to make it "official" for 1.83 and change it?

user-guide/modules/ROOT/pages/getting-started.adoc
```diff
@@ -0,0 +1,886 @@
...
+[source]
+----
+./bootstrap.sh
+./b2
+./b2 install --prefix=/usr/local
+----
+
+You can adjust the installation `prefix` to your preference in this command.
+
+NOTE:: We break the build and install into separate steps, in order to get better diagnostics when building the libraries, and to suppress too many messages when installing the libraries.
```
anarthal, line 240: We're mentioning breaking the build into two steps becaue of `-d0` but there's no `-d0` in any of the commands

user-guide/modules/ROOT/pages/getting-started.adoc
```diff
@@ -0,0 +1,886 @@
...
+====
+Boost.Build::
++
+--
+The Boost libraries are built using a custom build app called boost:build[].
+This app itself is built by running the bootstrap script.
+
+This is the usual procedure to configure, build and install Boost with boost:build[]:
+
+Decide where you are going to install the libraries to, most locations are OK, something like *Documents/Boost-libraries* is fine.
```
anarthal, line 212: shouldn't the commands use this? Or remove it alltogether, I think our users are old enough to know this :)

user-guide/modules/ROOT/pages/getting-started.adoc
```diff
@@ -0,0 +1,886 @@
...
+cmake --build . --target install --config Release
+----
+
+Although both the boost:build[] and the CMake scripts work fine, the final installed libraries will have different layouts for historical reasons.
+Both the CMake integration scripts and the libraries will be installed under different paths and this can create incompatibilities between Boost installations.
+
+TIP: You don't need to *build* Boost with CMake to be able to *use* Boost with CMake.
+CMake integration will work fine even if you install Boost with boost:build[].
+
+The supported and tested way to build Boost remains with boost:build[] and you should refrain from using CMake at this step unless you need this functionality.
```
anarthal, line 280: What does "this functionality" mean in this context? If we're presenting this as something advanced, it shouldn't be in the getting started, in my opinion.

user-guide/modules/ROOT/pages/getting-started.adoc
```diff
@@ -0,0 +1,886 @@
...
+target_link_libraries(MyProject Boost::headers)
+----
+
+Store this file next to your `example.cpp` and run the basic CMake workflow:
+
+[source,none,subs="attributes+"]
+----
+mkdir build <1>
+cd build <2>
+cmake .. -D BOOST_ROOT=path/to/{release_basename} <3>
```
anarthal, line 402: Does `BOOST_ROOT` work in the general case? I tend to prefer `CMAKE_PREFIX_PATH`, since it's not Boost-specific (you can consume anything by using it).

user-guide/modules/ROOT/pages/getting-started.adoc
```diff
@@ -0,0 +1,886 @@
...
+
+=== Header-only libraries
+
+Let’s start by using a header-only library.
+Copy the following program into a file called `example.cpp`.
+
+[source,C++]
+.example.cpp
+----
+#include <boost/lambda/lambda.hpp> <1>
```
anarthal, line 357: I don't know if this PR is the place to put this comment - but I'd provide another example. This is joining lambda (seems an old lib, and I can hear users "but we already have this in the language already") with iostreams (which we all love).

user-guide/modules/ROOT/pages/getting-started.adoc
```diff
@@ -0,0 +1,886 @@
...
+Did you get the expected result?
+
+=== Compiled Libraries
+
+In this section, we will locate an example from GitHub, and then build and run a project based on the example.
+
+For most Boost libraries, there is an `example` sub-folder containing a range of examples.
+For this guide, we will locate an example that reads a JSON file, and pretty-prints its contents.
+
+Copy all the code from https://github.com/boostorg/json/blob/develop/example/pretty.cpp[`json/example/pretty.cpp`,window="_blank"] into your `example.cpp`.
```
anarthal, line 545: While this is nice, if you dumb-follow the instructions, this won't work, since the example includes `"file.hpp"`.

user-guide/modules/ROOT/pages/getting-started.adoc
```diff
@@ -0,0 +1,886 @@
...
+./example "path/to/json/file.json"
+----
+--
+====
+
+Did you get the expected result?
+
+=== Summary
+
+Although the samples you have now built and run are quite simple, if you have got this far successfully, it means your build and installation and project linking are all working correctly.
```
anarthal, line 671: shouldn't be "build, installation and project linking"?

user-guide/modules/ROOT/pages/getting-started.adoc
```diff
@@ -0,0 +1,886 @@
...
+In this guide, we will discuss two types of package managers: pass:[C++] package managers such as Conan and Vcpkg, and system package managers such as Homebrew and APT.
+
+While using a package manager can simplify the installation process, there are some potential downsides to consider, such as a lack of control over the installation process and potential version conflicts with other installed software.
+We will discuss these in more detail below.
+
+=== pass:[C++] Package Managers
+
+pass:[C++] package managers are a popular choice for managing dependencies in pass:[C++] projects.
+They provide a simple and consistent way to download, configure, and install libraries and their dependencies that's replicable on various platforms.
+However, these installation methods are not officially supported, which means the packages are not regularly tested by Boost authors.
```
anarthal, line 692: I think this discourages too much the use of vcpkg (which I think it's "almost official") and conan. I think these are good to use, the ones that get you really old software are the Linux ones.

user-guide/modules/ROOT/pages/getting-started.adoc
```diff
@@ -0,0 +1,886 @@
...
+Another downside is the possibility of missing or outdated packages.
+
+==== Vcpkg
```
anarthal, line 695: Maybe having vcpkg & conan in tabs would be cool?

user-guide/modules/ROOT/pages/getting-started.adoc
```diff
@@ -0,0 +1,886 @@
...
+Update your package list to ensure you're getting the most recent version of the Boost libraries available in the repositories.
+
+[tabs]
+========
+Ubuntu::
++
+--
+To install the Boost pass:[C++] libraries onto Debian-based systems (such as Debian, Ubuntu), you can use the package manager apt.
+
+[source]
```
anarthal, line 746: I'd merge the `sudo apt update` and `sudo apt install libboost-all-dev` tabs in a single thing.

user-guide/modules/ROOT/pages/getting-started.adoc
```diff
@@ -0,0 +1,886 @@
...
+To verify that the Boost libraries have been installed correctly, you can check the version number:
+
+[source]
+----
+cat /usr/include/boost/version.hpp | grep "BOOST_LIB_VERSION"
+----
+
+This command should output the Boost version installed on your system.
+
+==== Homebrew
```
anarthal, line 851: would it make sense to put this into the tabs, too? Looks at the same level as the `apt`/`dnf` commands

anarthal (changes requested):

user-guide/modules/ROOT/pages/getting-started.adoc
alandefreitas: > I'm not sure whether we're really prepared to include cmake at this point

Yes. Your comments totally make sense. I'm taking this into account and will talk about different layouts in the docs. But we're working on cppalliance/boost which will contain (it already does) a proper cmake release so I'm looking ahead here. The only limitation ends up being the source layout (because of how libraries are structured), which doesn't really matter that much. The installed layout will work fine.

user-guide/modules/ROOT/pages/getting-started.adoc
alandefreitas: Yes. :laughing: I also thought about that. But I was just merging the content we had from 5 guides we had and concluded I would never finish this if I kept fixing the content. There are many weird sentences like that.

user-guide/modules/ROOT/pages/getting-started.adoc
alandefreitas: Yes. BOOST_ROOT tends to work even though you don't need it and there are many other ways. But setting the environment just gets it for free.

user-guide/modules/ROOT/pages/getting-started.adoc
alandefreitas: This is the example we have in the getting started page now. But I agree. I would open an issue for that because that requires some extra thinking about newer libraries we could use there. I think it would be nice to have something they feel they can only get from boost, like boost.unordered or something. Although boost.unordered doesn't have the same appeal to someone just getting started. Lambda2 might be a good option, although "2" in the first app doesn't look good to me. We definitely don't want something we have in the standard library. Well... it requires some thinking.

user-guide/modules/ROOT/pages/getting-started.adoc
alandefreitas: Yes. Same as the other example. We have to look for a good one. And test it in all dumb-follow paths.

PR #62, opened 2023-05-10
antora ui updates

novoio:


alandefreitas (approved):
That's great. I would make the commit msg a little less generic though.

@novoio please let me know when you think this is good to go

PR #77, opened 2023-05-12
style guide updates

novoio:


alandefreitas (changes requested):
We should remove this stuff from the root directory.

novoio:
> We should remove this stuff from the root directory.

Moved to new directory `style-guidelines`.

PR #81, opened 2023-05-15
search box moved to top left ToC area

novoio:


alandefreitas:
@vinniefalco  is this what you were expecting?

PR #91, opened 2023-05-17
fix edit this page broken link

novoio:


site.playbook.yml
```diff
@@ -35,10 +35,13 @@ content:
   sources:
     - url: https://github.com/cppalliance/site-docs
       start_path: user-guide
+      edit_url: '{web_url}/edit/${page-boost-branch}/{path}'
```
alandefreitas, line 38: I don't know where {web_url} is coming from, but the branch should always be develop.

.github/workflows/publish.yml
```diff
@@ -60,7 +60,7 @@ jobs:
           fi
           
           # Always build sitedoc because all other changes affect it
-          ./sitedoc.sh ${{ github.ref_name }}
+          ./sitedoc.sh ${{ github.event.pull_request.base.ref }}
```
alandefreitas, line 63: You probably don't want to mess with that. So many things can break when we change this (revsys, previews, tests, etc...).

novoio: @alandefreitas `${{ github.ref_name }}` will not work on pull requests. As you can see in below image it hold pull reference instead of branch reference. After I change it to `${{ github.event.pull_request.base.ref }}` now we can get target branch `develop` but `Edit this Page` link is still referring to `HEAD` branch IDK why. It's working fine on my local.
image

site.playbook.yml
novoio: @alandefreitas `web_url ` holds our repo link in antora ui.

.github/workflows/publish.yml
novoio: After this workflow change you can see the correct branch name.
image

novoio: And obviously we need to test it that it will works as it in all cases.

novoio: @alandefreitas I have made some changes for redirecting to develop branch from Edit this Page link. See if it looks good to you.

alandefreitas: The link in the preview is now https://github.com/cppalliance/site-docs/edit/develop/user-guide/modules/ROOT/pages/index.adoc

Does it look different locally?

novoio: No! it redirects to develop locally too.

alandefreitas:
Great! Could you squash these commits?

PR #96, opened 2023-05-24
Search box styling

novoio:


alandefreitas:
Oh... that's nice. The old style was kind of irritating. Is this Lurn / Algolia combo box here to stay or is it just a test? I don't see how it can make sense from the POV of the user. I would imagine something like merging the results or just ignoring one of them. I don't know.

Three other things I've noticed are that 

1) the box height changes when we toogle lurn / algolia

image

image

2) the arrow is remaining after we finish the search

image

3) The content of the box is erased after we search with algolia and click somewhere else but not with lunr.

novoio:
@alandefreitas It's ready to merge.

alandefreitas:
> The content of the box is erased after we search with algolia and click somewhere else but not with lunr.

I thought we were going to have the opposite here. Not erasing in both cases.

>  It's ready to merge.

Great. Thanks. Please squash the commits. :)

PR #98, opened 2023-05-25
Links and tutorial clean up

PeterTurcan:
Tested the boost.test tutorial, made a few small changes. Minor clean up elsewhere.


contributor-guide/modules/ROOT/pages/index.adoc
```diff
@@ -12,5 +12,5 @@ Start by reading about xref:getting-involved.adoc[how to get involved in Boost].
 
 == See Also
 
-* https://docs.cppalliance.org/user-guide/index.html[User Guide]
-* https://docs.cppalliance.org/formal-reviews/index.html[Formal Reviews]
+* link:../user-guide/index.html[User Guide]
```
alandefreitas, line 15: We need to fix that with proper xrefs. Relative links going to parent paths might work but they can break very easily.

contributor-guide/modules/ROOT/pages/release-process.adoc
```diff
@@ -60,8 +60,5 @@ The release candidate is published to the Boost site. The *master* branch is ope
 +
 If issues are found with a release candidate that are important enough to address quickly (that is, before the next full public release), then a point release will be built when fixes are available and tested. This will not typically result in the *master* branch being closed to other check ins.
 
-== See Also
-
-* https://docs.cppalliance.org/user-guide/index.html[User Guide]
-* https://docs.cppalliance.org/contributor-guide/index.html[Contributor Guide]
-* https://docs.cppalliance.org/formal-reviews/index.html[Formal Reviews]
+For details of the Release Process that are pertinent to users, refer to the 
+link:../user-guide/release-process.html[User Guide Release Process].
```
alandefreitas, line 64: same

user-guide/modules/ROOT/pages/index.adoc
```diff
@@ -16,6 +16,5 @@ Start by reading the xref:intro.adoc[Introduction to Boost].
 
 == See Also
 
-* https://docs.cppalliance.org/contributor-guide/index.html[Contributor Guide]
-* https://docs.cppalliance.org/formal-reviews/index.html[Formal Reviews]
+* link:../contributor-guide/index.html[Contributor Guide]
```
alandefreitas, line 19: same

user-guide/modules/ROOT/pages/boost-test-tutorial.adoc
```diff
@@ -13,11 +13,11 @@ Let's get started with a trivial example.
 
 Here's a very simple example of a test suite with Boost.Test:
 
-. Use Visual Studio to create a C++ Console application, call it something like "StringRev".
```
alandefreitas, line 16: The tutorial is quite nice. I don't think most people are on Visual Studio though (although Vinnie is). A lot of people use VSCode, jetbrains, etc... and even Visual Studio itself is moving towards CMake, so these workflows are kind of unmaintained and are going to be deprecated at some point. 

I think the best option would be to create some tabs at these steps where we are setting things up with "CMake", "Visual Studio", "By Hand", like in the "getting started" page. 

If that's too much work for now and we only want to have one option, this second best alternative for a single option would be CMake, because it's the de facto build system for C++ at the moment. You can make local tests with CMake directly from VS so you don't even have to change your workflow.

user-guide/modules/ROOT/pages/boost-test-tutorial.adoc
```diff
@@ -13,11 +13,11 @@ Let's get started with a trivial example.
...
+. Use Visual Studio to create a C++ Console application, call it something like "Trivia".
 
-. In the project *Properties* for *C++/General*, locate *Additional Include Directories* and add the path to your Boost libraries. The path will be something like "C:\Users\Peter\Documents\Boost\boost_1_81_0".
+. In the project *Properties* for *C++/General*, locate *Additional Include Directories* and add the path to your Boost libraries. The path will be something like "C:\Users\<your path>\boost_1_81_0".
 
-. Then, still in *Properties*, but now for *Linker/General* add to the *Additional Library Directories* with the path to your Boost lib folder. This path will be something like "C:\Users\Peter\Documents\Boost\lib".
+. Then, still in *Properties*, but now for *Linker/General* add to the *Additional Library Directories* with the path to your Boost lib folder. This path will be something like "C:\Users\<your path>\lib".
```
alandefreitas, line 20: I believe the convention for paths is using "`", just like we do for code.

That is, `C:\Users\Peter\Documents\Boost\lib` (with "`" as quotes) instead of "C:\Users\Peter\Documents\Boost\lib".

user-guide/modules/ROOT/pages/reporting-issues.adoc
```diff
@@ -6,37 +6,24 @@
...
+* Describe the problem carefully, including steps required to reproduce it by a library maintainers.
 
+* Attach a *minimal* and *complete* program that reproduces the problem. Aside from helping the library maintainer fix the problem, you may find the bug in your own code, which can avoid a costly delay waiting for a response.
 
 
+== If You Have a Fix for the Bug
 
+After completing the steps above, please submit a patch or a pull request as described in https://svn.boost.org/trac/boost/wiki/StartModPatchAndPullReq:[the wiki "Getting Started with Patches and Pull Requests"].
 
 Copyright Aleksey Gurtovoy(/users/people/aleksey_gurtovoy.html) 2002
```
alandefreitas, line 28: This link is not rendered properly. I think I would also put this in a box, admonition, or quotation because it looks weird as part of the content.

user-guide/modules/ROOT/pages/reporting-issues.adoc
```diff
@@ -6,37 +6,24 @@
...
-1. If you know where to look in the source code, make sure
- the bug isn't already fixed in the latest sources. The most
- recent version of everything on the Boost web site is
- available from the https://github.com/boostorg/wiki/wiki/Getting-Started:[git repositories].
-2. https://github.com/boostorg/:[Search the issues]
- on GitHub to make sure we don't already know about the bug.
- If we do, you can add further information to an existing bug ticket.
-3. https://github.com/boostorg/:[Create a new issues]
- in the repository of the particular library of your interest.
+. If you know where to look in the source code, make sure the bug isn't already fixed in the latest sources. The most  recent version of everything on the Boost web site is available from the https://github.com/boostorg/wiki/wiki/Getting-Started:[git repositories].
```
alandefreitas, line 9: Most subsections in "Common Scenarios" are kind of weird. Ignoring other subsections for now, I think this one could be under "Community". I can't see a person who just got started reporting issues. Neither we want to incentivize that.

contributor-guide/modules/ROOT/pages/index.adoc
PeterTurcan: @alandefreitas I have tried many xref options and none have worked. Might need to get some support from somewhere to work out the correct link format.

user-guide/modules/ROOT/pages/boost-test-tutorial.adoc
PeterTurcan: @alandefreitas OK, perhaps a VSC and CMake option too, I'll investigate.

user-guide/modules/ROOT/pages/boost-test-tutorial.adoc
PeterTurcan: @alandefreitas OK, will change this.

user-guide/modules/ROOT/pages/reporting-issues.adoc
PeterTurcan: OK

user-guide/modules/ROOT/pages/reporting-issues.adoc
PeterTurcan: I will move to Community.

anarthal: Idk if this is thw story for it, but I think the wiki is super unmaintained so I'd refrain from linking to it.

contributor-guide/modules/ROOT/pages/index.adoc
alandefreitas: I just tried `xref:user-guide:ROOT:index.adoc[]` and it worked fine.

image

So we have the syntax `xref:version@component:module:file.adoc#fragment[text]`, where:

- `version` is empty and therefore implies we want the same version (this playbook has no versions)
- `component` is `user-guide`
- `module` is `ROOT` and it can't be implied because we defined the component and the `file` component can't be ambiguous
- `file.adoc` is `index.adoc`
- `fragment` is empty as usual because we want to refer to the whole page
- `text` is empty because we want it to come from the `adoc` attributes

Using xref 1) lets us get the title from the adoc attributes, 2) doesn't rely on circumstantial patterns in the build layout, 3) allows the asciidoc plugin to perform static analysis (including indicating errors, auto-completion, syntax-highlighting, etc), and 4) removes ".." segments, which are unsafe.

The complete syntax is in https://docs.antora.org/antora/latest/page/xref/

user-guide/modules/ROOT/pages/boost-test-tutorial.adoc
alandefreitas: When you create a project, VS lets you choose between a cmake project and a "legacy" VS solution, which is now considered just one possible cmake target. Just replicated the pattern from "Getting Started" to link the libraries in CMake and all the rest of the content should be exactly the same.

user-guide/modules/ROOT/pages/reporting-issues.adoc
alandefreitas: Oh... I didn't even notice the link. These two sentences seem unrelated and out of place.

> If you know where to look in the source code, make sure the bug isn't already fixed in the latest sources. 

This first sentence is about reporting issues. And yes. The real and de facto place to report issues nowadays is the corresponding github repository. It's the place where the library author will get a notification and is more likely to reply unless it's boost.asio in which case you're out of luck anyway :laughing:. A second best place is the mailing list, but it's not 1% as good as the github repository in my opinion. Issues in the ML have a tragedy of the commons in terms of replies and are even somewhat annoying sometimes.

> The most recent version of everything on the Boost web site is available from the https://github.com/boostorg/wiki/wiki/Getting-Started:[git repositories].

So this second sentence is about something else entirely. For now, we could just remove this sentence because it's simply false. Once our (initial) work is done with site-docs, the website content will come from whatever becomes of this very repository and cppalliance/temp-site.

contributor-guide/modules/ROOT/pages/index.adoc
PeterTurcan: works, thanks, done.

PeterTurcan:
@alandefreitas Updated based on feedback (links, reporting issues). I will revisit the tutorial later after giving it more thought. I also added a FAQ.

alandefreitas:
OK. Please squash the commits then. :)

PeterTurcan:
@alandefreitas squashed

PR #99, opened 2023-05-26
Common scenarios populated, FAQ moved

PeterTurcan:
Added five common development scenarios (database, networking, simulation, system, text processing) and a useful selection of libs for them. Moved FAQ to top level.

### DIFF
--- a/user-guide/modules/ROOT/nav.adoc
+++ b/user-guide/modules/ROOT/nav.adoc
@@ -1,9 +1,13 @@
 * xref:intro.adoc[]
 * xref:whats-inside.adoc[]
 * xref:getting-started.adoc[]
+* xref:faq.adoc[]
 
 * Common Scenarios
-** xref:faq.adoc[]
+** xref:task-networking.adoc[]
+** xref:task-simulation.adoc[]
+** xref:task-system.adoc[]
+** xref:task-text-processing.adoc[]
 ** xref:testing-debugging.adoc[]
 
 * Tutorials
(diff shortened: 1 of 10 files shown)


alandefreitas:
I found it weird when the title said "Database" but didn't mention Boost.MySql when I clicked on it. Boost.Container being the first library in the list was kind of weird because that's quite generic and can be used in any kind of application. It seems like these list are not ordered by their importance to the application.

"Testing and Debugging" seems to be moving out of the common pattern of listing libraries now. It could list some useful libraries to make it uniform and could be included after "Text Processing" to make signal this section is slightly more special.

anarthal:
"Database" as in "writing yourself a database engine"? I don't think that is a common use case at all.

Networking should list Boost.MySQL and Boost.Redis.

PeterTurcan:
@alandefreitas @anarthal Changes made based on your feedback. I could not get boost:redis[] to work? - so I just linked to its github site. Moved database to "Advanced Scenarios" - and added Machine Learning, Metaprogramming, and Natural Language Parsing (stubs only) topics to advanced, if that makes sense.

alandefreitas:
> I could not get boost:redis[] to work? 

Yes. I shouldn't work because it hasn't been released yet.

> Moved database to "Advanced Scenarios" - and added Machine Learning, Metaprogramming, and Natural Language Parsing (stubs only) topics to advanced, if that makes sense.

Sure. If you have something in mind for these sections, the placeholders do look OK. 

> Changes made based on your feedback. 

Great. Please squash the commits when you think it's good to go.

PeterTurcan:
@alandefreitas squashed!

PR #103, opened 2023-06-07
Update README.adoc

4down:
sitedoc should be libdoc in example


4down:
Tailwindcss sets all un-classed lists to have a list-style of none. Antora UI  was assuming the CSS default of disc would remain unaltered.

alandefreitas:
> sitedoc should be libdoc in example

mmm ... makes sense

PR #109, opened 2023-06-12
Library Metadata section updated

PeterTurcan:
Checked and re-worked the Library Metadata section. Added new Json fields where found. Updated categories. Fixed a broken link in Getting Involved.

### DIFF
--- a/contributor-guide/modules/ROOT/pages/design-guide/dependencies.adoc
+++ b/contributor-guide/modules/ROOT/pages/design-guide/dependencies.adoc
@@ -3,74 +3,53 @@
 :idprefix:
 :idseparator: -
 
-*Legacy content, currently being updated.*
+A Boost library _does not generally_ depend on other libraries, except for Boost or the pass:[C++] Standard Library. There are some exceptions. For example, boost:iostreams[], and boost:beast[], have an optional dependency on https://www.zlib.net/[*zlib*], to provide support for gzip and zlib compression and decompression. If zlib is not available, boost:iostreams[] will still work, but without the ability to handle gzip and zlib formats. Another example is boost:asio[], which has an optional dependency on https://www.openssl.org/[*OpenSSL*], to provide support for SSL and TLS protocols. These protocols are widely used for secure network communications. Again, if OpenSSL is not available, boost:asio[] will still work, but without support for SSL or TLS. 
 
-A Boost library **should not** use libraries
- other than Boost or the C++ Standard Library.
+Other examples include boost:python[] interfaces with the Python C API, so obviously depends on a Python installation. boost:locale[] depends on the International Components for Unicode (https://icu.unicode.org/[*ICU*]) to provide Unicode and localization support. On Unix systems, boost:thread[] relies on the https://www.cs.cmu.edu/afs/cs/academic/class/15492-f07/www/pthreads.html[*pthread*] library. boost:compute[] has a dependency on https://www.khronos.org/opencl/[*OpenCL*].
 
+For other libraries that you might want to take a dependency on, get a discussion going first with the https://lists.boost.org/mailman/listinfo.cgi/boost[Boost developers' mailing list]. The general rule is avoid _unreasonable_ dependencies.
 
-A Boost library **should** use other Boost
- Libraries or the C++ Standard Library, but only when the
- benefits outweigh the costs.
+== Benefits and Effects of Coupling
 
+A Boost library _should_ use other Boost Libraries, or the pass:[C++] Standard Library, when the benefits outweigh the costs. In general, Boost libraries are designed to be as independent as possible, so that users can pick and choose the libraries they need without being forced to include unnecessary code. Many Boost libraries are indeed standalone and can be used separately without any dependencies on other libraries. 
 
-The benefits of using components from other libraries may
- include clearer, more understandable code, reduced development
- and maintenance costs, and the assurance which comes from
- reusing well-known and trusted building blocks.
+The benefits of using components from other libraries may include clearer, more understandable code, reduced development and maintenance costs, and the assurance which comes from reusing well-known and trusted building blocks. The costs may include undesirable coupling between components, and added compilation and runtime costs. If the interface to the additional component is complex, using it may make code less readable, and thus actually increase development and maintenance costs.
 
+Negative effects of coupling become obvious when one library uses a second library which uses a third, and so on. The worst form of coupling requires the user understand each of the coupled libraries. Coupling may also reduce the portability of a library - even in cases when all used libraries are self-sufficient (see <<Examples of Questionable Dependencies>>).
 
-The costs may include undesirable coupling between
- components, and added compilation and runtime costs. If the
- interface to the additional component is complex, using it may
- make code less readable, and thus actually increase development
- and maintenance costs.
+== Examples of Acceptable Dependencies
 
+The boost:graph[] library depends on boost:iterator[] and boost:property-map[]. Similarly, boost:asio[], mentioned above, depends on several other Boost libraries including boost:system[], boost:date-time[], and boost:bind[], and others. boost:asio[] itself is recommended if your library needs networking. 
 
-Negative effects of coupling become obvious when one library
- uses a second library which uses a third, and so on. The worst
- form of coupling requires the user understand each of the
- coupled libraries. Coupling may also reduce the portability of
- a library - even in case when all used libraries are
- self-sufficient (see example of questionable usage of
- <iostream> library below).
+A good example where another boost component should certainly be used is boost:core[], as it has considerable benefits; it simplifies code, improves readability, and signals intent. boost:core[] contains small utilities, usually polyfills for standard components, which you use when targeting pass:[C++] standards where they are not available. Costs are low as coupling is limited; the boost:core[] header includes only lightweight headers. There are no runtime costs at all. With costs so low and benefits so high, other boost libraries should use boost:core[] when the need arises.
 
+Other Boost libraries you might consider as good foundational components include:
 
-**Example where another boost component should
- certainly be used:** boost::noncopyable (in [boost/utility.hpp](/doc/libs/release/boost/utility.hpp))
- has considerable benefits; it simplifies code, improves
- readability, and signals intent. Costs are low as coupling is
- limited; noncopyable itself uses no other classes and its
- header includes only the lightweight headers
- <boost/config.hpp> and <cstddef>. There are no
- runtime costs at all. With costs so low and benefits so high,
- other boost libraries should use boost::noncopyable when the
- need arises except in exceptional circumstances.
+* boost:compatibility[] has been updated recently and has a similar role to boost:core[]. Its' classes provide polyfills from std classes, and have no extensions.
 
+* boost:config[] is used by almost all libraries. It provides macro definitions to make your code portable. It enables you to detect standard levels prior to pass:[C++]17, to check which platform/compiler you are building for, and to add the relevant platform-specific code to create compiled libraries.
 
-**Example where a standard library component might
- possibly be used:** Providing diagnostic output as a
- debugging aid can be a nice feature for a library. Yet using
- Standard Library <iostream> can involve a lot of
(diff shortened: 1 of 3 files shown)


PeterTurcan:
@anarthal @alandefreitas Updated Dependencies section added to this PR. Let me know if you have any feedback on either of the updates - thanks!

contributor-guide/modules/ROOT/pages/design-guide/dependencies.adoc
```diff
@@ -3,74 +3,46 @@
 :idprefix:
 :idseparator: -
 
-*Legacy content, currently being updated.*
+A Boost library _should generally not_ depend on other libraries, except for Boost or the pass:[C++] Standard Library. There are some exceptions. For example, boost:iostreams[] has an optional dependency on https://www.zlib.net/[*zlib*], to provide support for gzip and zlib compression and decompression. If zlib is not available, boost:iostreams[] will still work, but without the ability to handle gzip and zlib formats. Another example is boost:asio[], which has an optional dependency on https://www.openssl.org/[*OpenSSL*], to provide support for SSL and TLS protocols. These protocols are widely used for secure network communications. Again, if OpenSSL is not available, boost:asio[] will still work, but without support for SSL or TLS.
```
alandefreitas, line 6: >  _should generally not_

This is a little controversial because there's no underlying rule saying they can't or shouldn't depend on other libraries. For instance, Asio, which is one of the big ones, depends on OpenSSL for TSL. Boost.Compute depends on OpenCL. Beast depends on zlib for compressed HTTP communication. 

The underlying rule is that they should not have unreasonable dependencies because that would make things worse for the user and testing much more difficult. For instance, a library shouldn't depend on nlohmann/json when it can depend on boostorg/json, which boost users will already have. Any GUI library proposal would have dependencies outside boost. A library proposal for a new container type is unlikely to have extra dependencies. Dependencies also tend to be optional to avoid breaking the super-project build, although I'm not sure I completely agree with this rule.

In practice, this means most libraries end up with no dependencies outside other boost libraries. But this is a description of how things are, not how things should be unconditionally and we don't want to give this impression to prospective contributors. I'd rather describe that "A Boost library _does not generally_ depend on other libraries"...

contributor-guide/modules/ROOT/pages/design-guide/dependencies.adoc
```diff
@@ -3,74 +3,46 @@
...
 
-A Boost library **should not** use libraries
- other than Boost or the C++ Standard Library.
+Taking a dependency on https://www.zlib.net/[*zlib*] or https://www.openssl.org/[*OpenSSL*] is fine. boost:python[] interfaces with the Python C API, so obviously depends on a Python installation. boost:locale[] depends on the International Components for Unicode (https://icu.unicode.org/[*ICU*]) to provide Unicode and localization support. On Unix systems, boost:thread[] replies on the https://www.cs.cmu.edu/afs/cs/academic/class/15492-f07/www/pthreads.html[*pthread*] library.
```
alandefreitas, line 8: I think I ended up explaining the motivation above, which is basically no unnecessary dependencies.

contributor-guide/modules/ROOT/pages/design-guide/dependencies.adoc
```diff
@@ -3,74 +3,46 @@
...
-**Example where another boost component should not be
- used:** The boost dir\_it library has considerable
- coupling and runtime costs, not to mention portability issues
- for unsupported operating systems. While completely appropriate
- when directory iteration is required, it would not be
- reasonable for another boost library to use dir\_it just to
- check that a file is available before opening. C++ Standard
- Library file open functionality does this at lower cost. Don't
- use dir\_it just for the sake of using a boost library.
+* For handling data types and structures, one of boost:function-types[], boost:fusion[], boost:any[], boost:variant[] might provide what you need.
```
alandefreitas, line 35: boost::variant2

contributor-guide/modules/ROOT/pages/design-guide/dependencies.adoc
```diff
@@ -3,74 +3,46 @@
...
 
+* For metaprogramming, boost:mpl[] or boost:preprocessor[] provide many of the building blocks for this style of programming.
```
alandefreitas, line 37: boost::mp11 instead

contributor-guide/modules/ROOT/pages/design-guide/dependencies.adoc
```diff
@@ -3,74 +3,46 @@
...
 
+== Examples of Questionable Dependencies
+
+An example where a standard library component might be used, but with care, is in providing diagnostic output. This output is used as a debugging aid and can be a nice feature for a library. Yet using Standard Library `<iostream>` can involve a lot of additional cost, particularly if `<iostream>` is unlikely to be used elsewhere in the library users' application. In certain GUI or embedded applications, coupling to `<iostream>` would be a disqualification. In this case, consider a redesign of the boost library in question so that the user supplies the diagnostic output mechanism.
```
alandefreitas, line 41: > An example where a standard library component might be used, but with care

We don't have particular restrictions against "standard library components" at all in boost. We can't make it sound like being allowed to use the standard library is an exception.

Some standard headers should be avoided for multiple reasons in C++ but these reasons are not particular to Boost.

contributor-guide/modules/ROOT/pages/getting-involved.adoc
```diff
@@ -53,7 +53,7 @@ All libraries are reviewed by the community before becoming part of Boost. There
 
 If you have a deep interest in a particular library that for some reason no longer has an active owner or maintainer, or is being less well maintained than might be ideal (find this out by asking questions on the https://lists.boost.org/mailman/listinfo.cgi/boost[Boost developers' mailing list]), then consider requesting, again via the mailing list, to become a library maintainer. For larger libraries there can certainly be more than one maintainer.
```
alandefreitas, line 54: It's worth talking to the current maintainer listed in `meta.json` file before.

contributor-guide/modules/ROOT/pages/getting-involved.adoc
```diff
@@ -53,7 +53,7 @@ All libraries are reviewed by the community before becoming part of Boost. There
...
 
-Maintaining an existing library is similar in steps to <<3. Contribute to an Existing Library>>, in that you will have to create a fork and clone the library. However, more is expected. You will have to spend time educating yourself on the full functionality of the library, and be prepared to make changes to the code anywhere changes are needed. 
+Maintaining an existing library is similar in steps to <<Contribute to an Existing Library>>, in that you will have to create a fork and clone the library. However, more is expected. You will have to spend time educating yourself on the full functionality of the library, and be prepared to make changes to the code anywhere changes are needed. 
 
 For a larger library this education can turn out to be a significant, and perhaps onerous task. It can be quite difficult to reverse engineer in your mind the processes and purpose of code written by another developer, with all the expected personal style and idiosyncrasies that make their way into source code. However, that being said, it can be a valuable contribution to take an existing but deprecated library you are interested in and give it a new lease of life.
```
alandefreitas, line 58: > deprecated library

Libraries are not really deprecated. They might not be properly maintained (new features, ...) but not deprecated. They are all tested when the super-project is tested.

contributor-guide/modules/ROOT/pages/requirements/library-metadata.adoc
```diff
@@ -1,197 +1,189 @@
...
-System
-Miscellaneous
-Miscellaneous
-workarounds
-Broken compiler workarounds
+* 23 = pass:[C++]23
+
+The level only indicates the minimum level, which means that the functionality of the library can be used when compiling at that level or at a higher level. There may be some functionality in the library which will need a higher pass:[C++] standard compilation level than is indicated by this value, but the information about that specific functionality will be provided for the end-user within the documentation for that library. If a library does not have this field it indicates that the end-user will have to read the library documentation to understand what pass:[C++] standard compilation level is needed to use the library. 
+
+Note that `03`, `11`, and `14` are commonly set minimum levels.
```
alandefreitas, line 149: This is correct but it might be considered slightly prescriptive for new contributors, which is problematic since boost is no longer supporting C++03 and most libraries are moving to C++11: https://www.boost.org/users/history/version_1_82_0.html

In fact, we could just mention 11 and 14 here since they will already have deprecated C++03 at the time this documentation is online.

alandefreitas (commented):
Much information here is very dynamic. Basically which are the most common or possible values in `meta`. I worry about these pages becoming outdated and unmaintained in the future because we forgot to update them. At some point, maybe this information could come from a script or something more dynamic that generates antora partials.

PeterTurcan:
@alandefreitas @anarthal Good feedack - working on the updates and I'll try and get the PR ready by EOD

contributor-guide/modules/ROOT/pages/design-guide/dependencies.adoc
PeterTurcan: OK, reworded the paragraph

contributor-guide/modules/ROOT/pages/design-guide/dependencies.adoc
PeterTurcan: added

contributor-guide/modules/ROOT/pages/design-guide/dependencies.adoc
PeterTurcan: OK, replaced with mp11

contributor-guide/modules/ROOT/pages/design-guide/dependencies.adoc
PeterTurcan: OK, removed the reference to <iostream>

contributor-guide/modules/ROOT/pages/getting-involved.adoc
PeterTurcan: added that point

contributor-guide/modules/ROOT/pages/getting-involved.adoc
PeterTurcan: OK, got it, reworded

contributor-guide/modules/ROOT/pages/requirements/library-metadata.adoc
PeterTurcan: removed 03

contributor-guide/modules/ROOT/pages/design-guide/dependencies.adoc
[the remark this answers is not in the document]
PeterTurcan: good catch, fixed

contributor-guide/modules/ROOT/pages/design-guide/dependencies.adoc
```diff
@@ -3,74 +3,46 @@
...
- Standard Library <iostream> can involve a lot of
- additional cost, particularly if <iostream> is unlikely
- to be used elsewhere in the application. In certain GUI or
- embedded applications, coupling to <iostream> would be a
- disqualification. Consider redesign of the boost library in
- question so that the user supplies the diagnostic output
- mechanism.
+* boost:concept-check[] provides tools for specifying and checking that types meet the requirements of generic algorithms. It's used by many other libraries to ensure that template parameters meet the necessary requirements.
 
+* boost:type-traits[] provides tools for introspecting the properties of types, and is used extensively throughout Boost. It offers a collection of templates for querying and manipulating type information at compile-time.
```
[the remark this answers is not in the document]
PeterTurcan: removed type_traits as an example

contributor-guide/modules/ROOT/pages/design-guide/dependencies.adoc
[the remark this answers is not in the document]
PeterTurcan: Good input, worked this into the doc.

PeterTurcan:
@alandefreitas @anarthal I have made the updates based on your great feedback, and squashed to one commit. But let me know if you would like further changes or considerations.

PR #110, opened 2023-06-19
Explore the Content topic populated

PeterTurcan:
@alandefreitas @anarthal FYI

### DIFF
--- a/user-guide/modules/ROOT/pages/explore-the-content.adoc
+++ b/user-guide/modules/ROOT/pages/explore-the-content.adoc
@@ -1,0 +1,94 @@
+= Explore the Content
+:navtitle: Explore the Content
+
+There are a number of ways you can approach exploring Boost. Clearly, examining the documentation for any of the https://www.boost.org/doc/libs/1_82_0/libs/libraries.htm[Boost libraries] is a great way to start. However, with 180 or so libraries to choose from, you might want to consider a range of options to get going. This section covers a number of approaches you have to focus your exploration of the libraries, and get deeper into Boost content.
+
+If knowing what others have done before you helps, refer to <<High-Profile users of Boost>>. 
+
+If you are already using, or considering, the Standard Template Library then refer to <<Boost vs the Standard Library>> to get some idea of which to use when similar ones are available in both sets of libraries.
+
+If any of the libraries does pique your interest, consider examining and running working <<Examples>>. 
+
+There are mailing lists you might want to join too, to ask questions and join in the discussion, such as the https://lists.boost.org/mailman/listinfo.cgi/boost-users[Boost Users mailing list], or project-specific lists accessible from the https://www.boost.org/community/groups.html[Boost Mailing Lists] site.
+
+To delve into content for a specific scenario, refer to the *Common Scenarios* topics, such as:
+
+** xref:task-networking.adoc[]
+
+** xref:task-simulation.adoc[]
+
+** xref:testing-debugging.adoc[]
+
+And, to the *Advanced Scenarios* topics such as:
+
+** xref:task-database.adoc[]
+
+** xref:task-machine-learning.adoc[]
+
+** xref:task-parallel-computation.adoc[]
+
+== High-Profile users of Boost
+
+Some well known libraries and applications have used the building blocks that Boost provides. These include:
+
+* https://engineering.fb.com/2012/06/02/developer-tools/folly-the-facebook-open-source-library/[*Folly*], the Facebook Open Source Library, is an open-source library developed by Facebook that is intended to augment the Standard Library and Boost by providing additional features and optimizations, often with a focus on high-performance systems. It is a collection of relatively small and standalone libraries, and its features can be used on an as-needed basis.
++
+Some of these features include string formatting libraries, hash maps, dynamic types, and concurrency utilities. Folly makes extensive use of:
+
+    ** boost:context[] for creating and manipulating execution contexts, which are used in Folly's implementation of coroutines
+
+    ** boost:intrusive[] which provides containers that allow for the insertion and removal of elements in constant time
+
+    ** boost:optional[]  which provides a way to represent optional, or nullable, objects of a given type
++
+Additionally, many parts of Folly are designed to interact seamlessly with Boost and the Standard Library, making it easy to integrate Folly into existing codebases that use these libraries. Folly serves as an example of how Boost libraries can be used in the development of large-scale, high-performance systems. 
+
+* The https://bitcoin.org/en/bitcoin-core/[*Bitcoin Core*] client, which is the reference implementation of the Bitcoin protocol, makes extensive use of the Boost libraries. For example, it uses boost:asio[] for network programming, boost:thread[] for multithreading, and boost:test[] for unit tests.
+
+* https://www.mongodb.com/[*MongoDB*], a popular open-source document database, leverages libraries such as boost:filesystem[] for filesystem interactions, boost:thread[] for threading, and boost:smart_ptr[] for smart pointers.
+
+* The https://llvm.org/[*Low Level Virtual Machine*] (LLVM) project, which is a collection of modular and reusable compiler and toolchain technologies, utilizes Boost in parts of its code base. For instance, it uses boost:mpl[] for its metaprogramming functionality.
+
+NOTE: boost:mpl[] has now been superseded by boost:mp11[].
+
+* Microsoft has integrated boost:test[] into https://visualstudio.microsoft.com/[*Visual Studio*], in particular to enable unit testing.
+
+* Adobe also uses Boost in several of their software products. For instance, https://stlab.adobe.com/[*Adobe Source Libraries*] (ASL) includes and builds upon boost:smart_ptr[], boost:mpl[], boost:lambda[].
+
+* The https://www.ros.org/[*Robot Operating System*] (ROS), a flexible framework for robotics software development, uses Boost in several areas, for instance, boost:python[] to provide Python bindings, and boost:signals2[] for inter-process communication.
+
+* https://reactnative.dev/[*React Native*] is a popular JavaScript framework for building mobile application. Boost libraries are used extensively within JavaScriptCore (used to execute JavaScript within the app), and are a key part of the infrastructure that allows React Native to bridge between JavaScript and native code. The libraries used include boost:optional[], boost:intrusive[], and boost:multi-index[].
+
+* https://www.mysql.com/[*MySQL*] one of the most popular open-source relational database management systems, uses Boost libraries in its build process. It mainly uses the Boost header files, particularly those required for geometry processing operations, such as boost:geometry[].
+
+== Boost vs the Standard Library
+
+There are several Boost libraries that have a Standard Template Library (STL) equivalent but offer more features, better performance, or additional functionality that can make them more compelling in certain situations. When comparing libraries which have similar features, also consider compilation time, binary size, code readability, and the portability of your code.
+
+* boost:any[] is similar to `std::any`, introduced in pass:[C++]17. However, if you're working with a codebase that needs to be compatible with pre-pass:[C++]17 standards, then boost:any[] can be a lifesaver.
+
+* Similar to the situation with boost:any[], boost:optional[] provides functionality similar to `std::optional`, introduced in pass:[C++]17, but works with earlier standards.
+
+* While pass:[C++]11 introduced smart pointers into the standard library (with `std::unique_ptr` and `std::shared_ptr`), boost:smart_ptr[] offers some additional features. These features include the use of `intrusive_ptr` which allows for the customization of the reference-counting mechanism.
+
+* boost:container[] offers some advanced containers like `flat_map`, `small_vector`, `static_vector`, and `stable_vector` that are not available in STL `std::containers`.
+
+* boost:random[] has more extensive features than `std::random`.
+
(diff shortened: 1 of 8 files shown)


user-guide/modules/ROOT/pages/explore-the-content.adoc
```diff
@@ -0,0 +1,120 @@
...
+
+* The https://bitcoin.org/en/bitcoin-core/[*Bitcoin Core*] client, which is the reference implementation of the Bitcoin protocol, makes extensive use of the Boost libraries. For example, it uses boost:asio[] for network programming, boost:thread[] for multithreading, and boost:test[] for unit tests.
+
+* https://www.mongodb.com/[*MongoDB*], a popular open-source document database, leverages libraries such as boost:filesystem[] for filesystem interactions, boost:thread[] for threading, and boost:smart_ptr[] for smart pointers.
+
+* The https://llvm.org/[*Low Level Virtual Machine*] (LLVM) project, which is a collection of modular and reusable compiler and toolchain technologies, utilizes Boost in parts of its code base. For instance, it uses boost:mpl[] for its metaprogramming functionality.
+
+NOTE: boost:mpl[] has now been superseded by boost:mp11[].
+
+* Microsoft uses Boost in several projects, including in https://visualstudio.microsoft.com/[*Visual Studio*], which uses boost:test[] for some of its unit testing.
```
alandefreitas, line 39: How do we know about this? Is there some public link?

user-guide/modules/ROOT/pages/explore-the-content.adoc
```diff
@@ -0,0 +1,120 @@
...
+
+* Adobe also uses Boost in several of their software products. For instance, https://stlab.adobe.com/[*Adobe Source Libraries*] (ASL) includes and builds upon boost:smart_ptr[], boost.mpl[], boost:lambda[], and other libraries too.
```
alandefreitas, line 41: "and other libraries too" seems redundant since the sentence already starts with "For instance"

user-guide/modules/ROOT/pages/explore-the-content.adoc
```diff
@@ -0,0 +1,120 @@
...
+If knowing what others have done before you helps, refer to <<High-Profile users of Boost>>. If you like hard-copy format, there is a range of published <<Books>>. If you prefer an online format, there are quite a few tutorials published on <<YouTube>>. These options are good at giving the overall picture. 
+
+If you are already using, or considering, the Standard Template Library then refer to <<Boost vs the Standard Template Library>> to get some idea of which to use when similar ones are available in both sets of libraries.
+
+If you have very specific technical questions <<Stack Overflow>> may well have the answer. Alternatively, you can examine and run <<Examples>> for any of the libraries. 
+
+There are mailing lists you might want to join too, to ask questions and join in the discussion, such as the https://lists.boost.org/mailman/listinfo.cgi/boost-users[Boost Users mailing list], or project-specific lists accessible from the https://www.boost.org/community/groups.html[Boost Mailing Lists] site.
+
+
+== High-Profile users of Boost
```
anarthal, line 15: If I were a user, after reading this page, I would still not know what's in Boost (this may just be an opinion). It may be difficult summarizing all we have - but I'd say we should try (and maybe get more people involved, since "what's important" may be different for me and for others).

I'd try to mention networking (Asio and Beast are very widely used), maybe containers (a lot of work has recently gone into Unordered), URL, JSON, Geometry, Math and MP11, to note a few.

As I said, it's just an opinion. If you decide to go down the path I propose, more reviewers should look at this page before it gets submitted.

anarthal: I've just seen that this may be already covered in the "Common Scenarios" and "Advanced Scenarios", so my comment may make less sense. It may be worth considering placing a link to these sections here.

user-guide/modules/ROOT/pages/explore-the-content.adoc
```diff
@@ -0,0 +1,120 @@
...
+* https://www.amazon.com/Introduction-Boost-Libraries-II-Advanced/dp/9491028022/ref=asc_df_9491028022/?tag=hyprod-20&linkCode=df0&hvadid=266173100564&hvpos=&hvnetw=g&hvrand=11155108834785857002&hvpone=&hvptwo=&hvqmt=&hvdev=c&hvdvcmdl=&hvlocint=&hvlocphy=9033326&hvtargid=pla-571858518475&psc=1[Introduction to the Boost C++ Libraries; Volume II - Advanced Libraries], by Robert Demming and Daniel J. Duffy
+
+* https://www.amazon.com/Boost-Application-Development-Cookbook-application/dp/1787282244/ref=asc_df_1787282244/?tag=hyprod-20&linkCode=df0&hvadid=312280575053&hvpos=&hvnetw=g&hvrand=11155108834785857002&hvpone=&hvptwo=&hvqmt=&hvdev=c&hvdvcmdl=&hvlocint=&hvlocphy=9033326&hvtargid=pla-415478529500&psc=1[Boost C++ Application Development Cookbook - Second Edition: Recipes to simplify your application development], by Antony Polukhin
+
+
+== YouTube
+
+Consider searching YouTube for online tutorials:
+
+* https://www.youtube.com/results?search_query=Boost+library+tutorial[Boost Library Tutorial]
```
anarthal, line 73: I don't know if providing a link to a YouTube search adds enough value - I'd say most people know how to do it.

user-guide/modules/ROOT/pages/explore-the-content.adoc
PeterTurcan: Reworded this, as my sources are ambiguous. Just referring to the integration of boost:test into VS now.

user-guide/modules/ROOT/pages/explore-the-content.adoc
PeterTurcan: Agreed, removed the phrase.

user-guide/modules/ROOT/pages/explore-the-content.adoc
```diff
@@ -0,0 +1,120 @@
...
+
+== Boost vs the Standard Template Library
+
+There are several Boost libraries that have a Standard Template Library (STL) equivalent but offer more features, better performance, or additional functionality that can make them more compelling in certain situations. When comparing libraries which have similar features, also consider compilation time, binary size, code readability, and the portability of your code.
+
+* boost:any[] is similar to `std::any`, introduced in pass:[C++]17. However, if you're working with a codebase that needs to be compatible with pre-pass:[C++]17 standards, then boost:any[] can be a lifesaver.
+
+* Similar to the situation with boost:any[], boost:optional[] provides functionality similar to `std::optional`, introduced in pass:[C++]17, but works with earlier standards.
+
+* While pass:[C++]11 introduced smart pointers into the standard library (with `std::unique_ptr` and `std::shared_ptr`), boost:smart_ptr[] offers some additional features. These features include support for arrays, which `std::shared_ptr` does not enable, and the use of `intrusive_ptr` which allows for the customization of the reference-counting mechanism.
```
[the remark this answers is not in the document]
PeterTurcan: OK, removed the reference to arrays.

user-guide/modules/ROOT/pages/explore-the-content.adoc
```diff
@@ -0,0 +1,120 @@
...
+* https://www.amazon.com/Boost-Application-Development-Cookbook-application/dp/1787282244/ref=asc_df_1787282244/?tag=hyprod-20&linkCode=df0&hvadid=312280575053&hvpos=&hvnetw=g&hvrand=11155108834785857002&hvpone=&hvptwo=&hvqmt=&hvdev=c&hvdvcmdl=&hvlocint=&hvlocphy=9033326&hvtargid=pla-415478529500&psc=1[Boost C++ Application Development Cookbook - Second Edition: Recipes to simplify your application development], by Antony Polukhin
+
+
+== YouTube
+
+Consider searching YouTube for online tutorials:
+
+* https://www.youtube.com/results?search_query=Boost+library+tutorial[Boost Library Tutorial]
+
+== Boost vs the Standard Template Library
```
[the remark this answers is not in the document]
PeterTurcan: ok

user-guide/modules/ROOT/pages/explore-the-content.adoc
PeterTurcan: OK, added links to common and advanced scenarios

user-guide/modules/ROOT/pages/explore-the-content.adoc
PeterTurcan: OK, dropped the link and added a bit of detail on the content of the youtube tutorials.

PeterTurcan:
@alandefreitas @anarthal Updated Explore the Content from your feedback - and squashed to 1 commit.

alandefreitas:
Considering @anarthal 's comment

> If I were a user, after reading this page, I would still not know what's in Boost

I had a look at the hierarchy and I think we have a general problem with the hierarchy and 2 problems with this new page:

1) [Explore the Content](https://110.site-docs.prtest2.cppalliance.org/user-guide/explore-the-content.html) coming before [Getting Started](https://110.site-docs.prtest2.cppalliance.org/user-guide/getting-started.html) doesn't seem to be helping the average user of this documentation. I'd say most people visiting this documentation have a reasonable idea of what Boost is and just want to the snippets they need to copy/paste to install it. A discussion about who else is using it, books about boost, links, etc, is nice but not a requirement to get started at all. Making them go through that is not helping.

2) Rubéz's question only makes sense because the content of this page ended up being redundant with the "Common Scenarios". "Explore the Content" somewhat broke the natural hierarchy of things because ended up with two kinds of content: "High-Profile users of Boost" (which are use cases, like "Common Scenarios"), and "Resources about Boost" (which shouldn't be at the start because site-docs itself is the reference resource about boost users should be using at this point). The name is misleading because the content doesn't tell me how to "Explore the Content" in this documentation: it tells me some use cases and points me to some extra resources instead.

This is how I think these problems should be fixed:

1) "Common Scenarios" / "Tutorials" / "Advances Scenarios" breaks the flow. It should be "Common Scenarios" / "Advanced Scenarios" / "Tutorials". We could consider "Tutorials" / "Common Scenarios" / "Advances Scenarios" if we had more Tutorials.
2) "Advanced Builds" should be moved from "Advanced Scenarios" to right after "Getting Started" because it's not a "Use Case" after boost is built at all. It's just about non-default build configurations which might be necessary to get started on some platforms or even just use another compiler.
3) "Advanced Builds" gets renamed to "Build Configuration" / "Install Configuration" or something because "Advanced Builds" is unnecessarily scary.
4) "Common Scenarios" and "Advanced Scenarios" are renamed "Common Use Cases" and "Advanced Use Cases" to a) stop implying they're also about the configuration step or any other getting started step, b) make it clear they're about uses of Boost once it's installed.
5) Content about "High-Profile users of Boost" is merged with "Use Cases" because a) it's unrelated to exploring site-docs, and b) is not an external resource. The list of high-profile users can go into a new first page in "Common Use Cases". Specific uses of libraries by these high-profile users (for instance, why folly uses boost.context), go as admonitions in the corresponding page for that use case.
6) Sections about extra resources (Books, YouTube, Stack Overflow, etc...) go into a new "Resources" page before "Reference". The users trying to read the main reference resource about boost shouldn't be initially directed to *extra* resources.
7) Each library comes with its own tutorials, so I don't think we should have a Tutorials page unless we have a clear vision for what should be in it and why it's different from what the libraries can and do already provide. Unless something changes in that sense, I would put a "Tutorials" section in the "Resources" page, discuss how to find tutorials for each library, and call it a day. The current Boost.Test Tutorials would be removed from nav.adoc until we have this clear vision.

@anarthal  ?

PeterTurcan:
@anarthal Added a commit to address your issues on Metaprogramming and Database. I added a database client section, but if it doesnt make sense, I'll remove it. Updated the FAQ to refer to mp11, rather than mpl.

anarthal:
@PeterTurcan we already have the "networking" use case, which includes Redis and MySQL. I'd say to get rid of the "Database client" section entirely.

PeterTurcan:
@alandefreitas @anarthal Your comments all sound fine, I'll get a new commit with the changes in shortly. The point of the "high profile users" is I think mostly to provide users with the confidence that Boost is industrial strength, providing ideas I think less so - that is more the role of the common and advanced scenarios.

alandefreitas:
> I'd ask a question here first: what's the point of this page - give ideas to the user, or showcase important stakeholders we have to boost user confidence?

This is a good question. I don't know the answer but I guess if you ask people you're going to hear both answers. Most importantly I think the current content is trying to answer both questions with a paragraph about the user (folly is X and Y) and some examples about their use cases (folly uses boost.A to do B, boost.C to do D, etc...).

The initial paragraph about high-profile users is the one about inspiring confidence. It's not a use case but I don't know where the marketing to inspire confidence should go. In any case, I don't think it should be a requirement to "get started"...

After the first paragraph, the specific use cases could be admonitions in the use cases or just regular content.

PeterTurcan:
@alandefreitas @anarthal Updating commit includes:
Moved the books etc. stuff to a new Resource Page
Moved the test tutorial under the Testing and Debugging common scenario - perhaps all common scenarios should have a tutorial?
Left the high profile cases in the Explore Content for now - not yet sure of its right location - could be its own page
Adjusted the Contents order
Made the other suggestions (issues) on Networking, Database

If this works for now, I can squash and perhaps get some feedback

PR #123, opened 2023-07-24
Advanced Builds populated

PeterTurcan:
@anarthal Based party on your notes with some added scenarios. Currently it is in the Reference section, but could be moved.

FYI: @alandefreitas @fpelliccioni @sdkrystian


user-guide/modules/ROOT/pages/advanced-builds.adoc
```diff
@@ -1,3 +1,247 @@
 = Advanced Builds
+:navtitle: Advanced Builds
```
alandefreitas, line 2: I find it weird to call these "advanced builds"... probably because "advanced" might be a little ambiguous and with the connotation of being "hard", while these are basically just "build options" with nothing particularly hard about them.

user-guide/modules/ROOT/pages/advanced-builds.adoc
```diff
@@ -1,3 +1,247 @@
...
+
+This section covers a number of advanced build scenarios that you may come across if your project has complex, legacy, or non-standard, components or processes.
+
+== Build with a Non-default Compiler
+
+Using the `b2` build system, you can specify the compiler you want to use with the `toolset` option. For example, if you want to use the GNU Compiler Collection (GCC), you would specify `toolset=gcc`. If you want to use Clang, you would use `toolset=clang`. For example, to build the Boost libraries with gcc:
+
+[source,txt]
```
alandefreitas, line 10: These are not `txt`.

user-guide/modules/ROOT/pages/advanced-builds.adoc
```diff
@@ -1,3 +1,247 @@
...
+If the compiler is not in your system's PATH, you will need to provide the full path to the compiler executable. If there are spaces in the path, then wrap it in quotes, for example:
+
+[source,txt]
+----
+b2 toolset="C:\Program Files (x86)\Clang\"
+----
+
+== Specify the Application Binary Interface
+
+Building Boost with the old Application Binary Interface (ABI)) refers to compiling the library with a flag to ensure backward compatibility with an older version of an ABI.
```
alandefreitas, line 38: This subsection could move to the end. It's way more specific than the previous and next subsections. It's also not particularly related to boost because this is only a flag and how to set flags in general is the only thing related to boost b2 here. Setting the ABI is also way more specific than the use cases of other examples.

user-guide/modules/ROOT/pages/advanced-builds.adoc
```diff
@@ -1,3 +1,247 @@
...
+./b2 toolset=gcc cxxflags="-D_GLIBCXX_USE_CXX11_ABI=0"
+----
+
+This command specifies the compiled Boost libraries are to be linked with code that was also compiled with the old ABI. Setting the flag to 1 will specify the new ABI, though of course this is the default and the flag is unnecessary.
+
+Check compatibility with your entire toolchain and codebase when making decisions about which ABI to use. It's generally best to use the new ABI if all your code and dependencies support it.
+
+== Specify the Standard Library Implementation
+
+When using Boost libraries in a project, you might choose to use `pass:[libc++]` or `pass:[libstdc++]`. These are two different standard library implementations that you can use with your pass:[C++] compiler. The choice between the two depends on the compiler you are using, the platforms you are targeting, the standards and features you are using, and the specific requirements of your project.
```
alandefreitas, line 57: >  The choice between the two depends on the compiler you are using, the platforms you are targeting, the standards and features you are using, and the specific requirements of your project.

This is very vague.

user-guide/modules/ROOT/pages/advanced-builds.adoc
```diff
@@ -1,3 +1,247 @@
...
+[source,txt]
+----
+clang++ -stdlib=libc++ YOUR_FILE.cpp -lboost_system
+----
+
+If you're using GCC and you want to use  `pass:[libstdc++]`, you don't need to do anything specific because this is the default standard library for GCC. If you're using Clang and want to use  `pass:[libstdc++]`, you can use the `-stdlib=libstdc++` flag:
+
+[source,txt]
+----
+clang++ -stdlib=libstdc++ your_file.cpp -lboost_system
```
alandefreitas, line 76: The text mentions GCC and the example uses clang.

I'm not even sure the GCC flags are the same. 

I've only ever used `-stdlib` with clang to choose between libc++ and libstdc++ because clang provides libc++ but it defaults to libstdc++ on linux to avoid incompatibilities with GCC. I have never messed with that for any other reason and always use the default std implementation otherwise.

user-guide/modules/ROOT/pages/advanced-builds.adoc
```diff
@@ -1,3 +1,247 @@
...
+[source,txt]
+----
+./b2 toolset=clang cxxflags="-stdlib=libc++" linkflags="-stdlib=libc++"
+----
+
+
+
+== Specify Static or Dynamic Linking
+
+The compiler options `link=static` and `link=dynamic` specify how the Boost libraries should be linked to your project. If you do not specify a type of linking, `dynamic` is the default.
```
alandefreitas, line 90: > `dynamic` is the default.

Is that correct? When we build boost with the default instructions I get both static and dynamic libraries. When building individual libraries, we usually expect static linking by default.

user-guide/modules/ROOT/pages/advanced-builds.adoc
```diff
@@ -1,3 +1,247 @@
...
++
+[source,txt]
+----
+./b2 link=dynamic
+./b2 link=shared
+----
+
+Not all Boost libraries need to be linked to your program. Many are "header-only", meaning you only need to include their headers in your source files and don't need to worry about linking, as there is nothing to link. 
+
+== Build Boost with CMake
```
alandefreitas, line 111: We already have a section about building boost with cmake in the getting started guide. The user should already know that they can build with both b2 and cmake and that b2 is recommended. This section could be about cmake options, but it can't be about presenting the possibility of cmake because that already happened a few pages ago.

user-guide/modules/ROOT/pages/advanced-builds.adoc
```diff
@@ -1,3 +1,247 @@
...
+
+The only supported method of building the Boost libraries is with the `b2` tool. However, there are experimental alternatives, including building with CMake. Refer to the https://github.com/boostorg/cmake[Boost CMake support infrastructure] for full details, including information on general and library-specific configuration variables, testing after building with CMake, using Boost (or an individual Boost library) with the `add_subdirectory` command, and using Boost with the `FetchContent` command.
```
alandefreitas, line 113: > The only supported method of building the Boost libraries is with the `b2` too

b2 is recommended, but cmake is not unsupported. Peter has scripts that test it regularly.

user-guide/modules/ROOT/pages/advanced-builds.adoc
```diff
@@ -1,3 +1,247 @@
...
+
+== Parallel Builds
+
+To speed up the build process, you might want to utilize parallel compilation. This is where multiple compilation tasks are performed concurrently to speed up the build process. You can enable this feature by using the `-j` option followed by the number of tasks you want to run in parallel when invoking `b2`.
+
+For example, let's say you have a Jamroot.jam file at the root of your project that describes how to build your project. The project consists of several source files spread across different directories, and you are working on a multi-core machine where parallel builds could significantly speed up the build time. To build your project using Boost.Build, you would normally navigate to your project's root directory and run `b2`. However, to take advantage of parallel builds, you can specify the `-j` option followed by the number of tasks you want to run concurrently. If you have a quad-core processor, you might run up to four tasks in parallel:
+
+[source,txt]
+----
+b2 -j4
```
alandefreitas, line 123: I imagined we were changing the top level topic from `b2` to `cmake` here. If the intention was `cmake` instead of `b2`, then `cmake -j4` is not a valid command. If we moved the higher level topic from `b2` to `cmake` for just one paragraph and moved back to `b2` without listing any cmake build options, then the previous section is completely out of place.

user-guide/modules/ROOT/pages/advanced-builds.adoc
```diff
@@ -1,3 +1,247 @@
...
+----
+
+Running concurrent tasks can increase the system load and may not always result in a linear speedup. However, it is a great idea to use your multiple cores, if you have them available.
+
+== Custom Build Steps
+
+Your application might require pre-processing steps before compiling. For instance, you might need to generate some pass:[C++] code based on a configuration file. Custom build steps in Boost.Build can be achieved using _rule definitions_ and _actions_. Rules in Boost.Build are a little like functions in a programming language. You can define your own rule to execute custom commands, and then invoke the rule at the appropriate point in your Jamfile.
```
alandefreitas, line 130: I'm not even sure I understand this subsection. We went from something very basic (listing how to set build options when only INVOKING b2 to build boost, where the user doesn't have to know anything about b2 at all) to a very very VERY specific and very very VERY advanced topic for someone who's using b2 for their own application AND needs to create custom rules (few people in the world ever needed that) AND all before presenting any other way more basic b2 concepts (which is correct because this should be in the b2 documentation).

alandefreitas (changes requested):
You seem to be a little confused about the difference between 

1) invoking b2 to build boost and move on with your life and 
2) learning how to use b2 for your own projects

user-guide/modules/ROOT/pages/advanced-builds.adoc
anarthal: Agree.

user-guide/modules/ROOT/pages/advanced-builds.adoc
```diff
@@ -1,3 +1,247 @@
...
+[source,txt]
+----
+b2 toolset=clang
+----
+
+If the compiler is not in your system's PATH, you will need to provide the full path to the compiler executable. If there are spaces in the path, then wrap it in quotes, for example:
+
+[source,txt]
+----
+b2 toolset="C:\Program Files (x86)\Clang\"
```
anarthal, line 33: I was convinced this was not possible - did you check that this works?

user-guide/modules/ROOT/pages/advanced-builds.adoc
anarthal: No, that's not correct. This specifies how the libraries are built. You normally consume them through cmake. Under Linux, `link=static` builds only libboost_xxx.a, `link=shared` builds only `libboost_xxx.so`, and not specifying anything builds both. Windows works differently, probably. The only use case for this I can think of is reducing Boost build times if you know you won't need a specific variant of a library. Probably not that common, so we can move it down the document.

user-guide/modules/ROOT/pages/advanced-builds.adoc
```diff
@@ -1,3 +1,247 @@
...
+
+Not all Boost libraries need to be linked to your program. Many are "header-only", meaning you only need to include their headers in your source files and don't need to worry about linking, as there is nothing to link. 
+
+== Build Boost with CMake
+
+The only supported method of building the Boost libraries is with the `b2` tool. However, there are experimental alternatives, including building with CMake. Refer to the https://github.com/boostorg/cmake[Boost CMake support infrastructure] for full details, including information on general and library-specific configuration variables, testing after building with CMake, using Boost (or an individual Boost library) with the `add_subdirectory` command, and using Boost with the `FetchContent` command.
+
+== Parallel Builds
+
+To speed up the build process, you might want to utilize parallel compilation. This is where multiple compilation tasks are performed concurrently to speed up the build process. You can enable this feature by using the `-j` option followed by the number of tasks you want to run in parallel when invoking `b2`.
```
anarthal, line 117: Unlike `make`, `b2` already does this by default. If nothing is specified, b2 will use `jN` where N is the number of CPUs available in your machine.

user-guide/modules/ROOT/pages/advanced-builds.adoc
```diff
@@ -1,3 +1,247 @@
...
+
+For example, let's say you have a Jamroot.jam file at the root of your project that describes how to build your project. The project consists of several source files spread across different directories, and you are working on a multi-core machine where parallel builds could significantly speed up the build time. To build your project using Boost.Build, you would normally navigate to your project's root directory and run `b2`. However, to take advantage of parallel builds, you can specify the `-j` option followed by the number of tasks you want to run concurrently. If you have a quad-core processor, you might run up to four tasks in parallel:
```
anarthal, line 119: > For example, let's say you have a Jamroot.jam file at the root of your project that describes how to build your project.

This sentence is weird, since the user does not have a Jamroot.jam file in her project. They just build Boost using b2, not their projects.

user-guide/modules/ROOT/pages/advanced-builds.adoc
anarthal: No user is building their projects through b2. This does not make sense in the user guide.

user-guide/modules/ROOT/pages/advanced-builds.adoc
```diff
@@ -1,3 +1,247 @@
...
+== Multi-platform Support
+
+Let’s say you are developing an application that needs to run on multiple platforms, including Linux, Windows, and MacOS. Your app uses boost:asio[] for asynchronous networking, boost:thread[] for multithreading, and boost:filesystem[] for file handling:
+
+. Before you can use Boost libraries, they might need to be compiled. You can use Boost.Build to compile the libraries by navigating to the root directory of Boost and running `b2`. If you want to compile only the required libraries, use the `--with-` option:
+
++
+[source,txt]
+----
+./b2 --with-thread --with-filesystem --with-system
```
anarthal, line 176: This is really important and should be mentioned first, if not in the getting started.

user-guide/modules/ROOT/pages/advanced-builds.adoc
```diff
@@ -1,3 +1,247 @@
...
+
+lib boost_thread
+    : : <name>boost_thread
+      <search>path/to/boost/libs/thread/build
+    ;
+
+lib boost_filesystem
+    : : <name>boost_filesystem
+      <search>path/to/boost/libs/filesystem/build
+    ;
```
anarthal, line 200: Users don't write Jamfiles

user-guide/modules/ROOT/pages/advanced-builds.adoc
```diff
@@ -1,3 +1,247 @@
...
+----
+
+If you want to use a specific _version_ of a compiler, you can specify it after the compiler name, separated by a dash. For example, to use GCC 7.3, you would use `toolset=gcc-7.3`:
+
+[source,txt]
+----
+./b2 toolset=gcc-7.3
+----
+
+You can also specify the toolset option when you're building your application. This is done in a similar way, by passing `toolset=` to the `b2` command, as part of your build process. Here's an example:
```
sdkrystian, line 22: > Here's an example:
I would remove this and instead terminate the previous sentence with a colon

user-guide/modules/ROOT/pages/advanced-builds.adoc
```diff
@@ -1,3 +1,247 @@
...
+
+[source,txt]
+----
+b2 toolset=clang
+----
+
+If the compiler is not in your system's PATH, you will need to provide the full path to the compiler executable. If there are spaces in the path, then wrap it in quotes, for example:
```
sdkrystian, line 29: "Paths containing spaces must be enclosed in quotes:"

user-guide/modules/ROOT/pages/advanced-builds.adoc
sdkrystian, line 38: > (ABI))
extra parentheses here

user-guide/modules/ROOT/pages/advanced-builds.adoc
```diff
@@ -1,3 +1,247 @@
...
+[source,txt]
+----
+b2 toolset="C:\Program Files (x86)\Clang\"
+----
+
+== Specify the Application Binary Interface
+
+Building Boost with the old Application Binary Interface (ABI)) refers to compiling the library with a flag to ensure backward compatibility with an older version of an ABI.
+
+The ABI of a system is a specification that defines requirements for binary compatibility between applications and libraries that are built separately. It includes aspects such as calling conventions, byte order, and data structure alignment.
```
sdkrystian, line 40: I would use "endianness" instead of "byte order", but I suppose either works

user-guide/modules/ROOT/pages/advanced-builds.adoc
```diff
@@ -1,3 +1,247 @@
...
+
+In the context of pass:[C++], this generally comes up with GCC 5.x and later versions, which introduced a new ABI to conform more closely with the pass:[C++11] and pass:[C++14] standards. This new ABI changed the implementations of `std::string` and `std::list`, among other things, which broke binary compatibility with older versions of these libraries. This means that code compiled with the new ABI may _not_ be binary-compatible with code compiled with the old ABI.
+
+To build Boost with the old ABI when using GCC, you'd pass in the `-D_GLIBCXX_USE_CXX11_ABI=0` flag. For example:
```
sdkrystian, line 44: "Boost can be built with the old ABI using the `-D_GLIBCXX_USE_CXX11_ABI=0` flag:"

Since we already explain toolsets earlier, and because libstdc++ can be used with other compilers, I don't think its necessary to specify which particular compiler is used.

user-guide/modules/ROOT/pages/advanced-builds.adoc
```diff
@@ -1,3 +1,247 @@
...
+
+. `pass:[libc++]` is the standard library implementation developed by the https://github.com/llvm/llvm-project[LLVM project]. It is the default library for the Clang compiler and is known for its performance and standards compliance. It's also designed to work well with the Clang static analyzer and other LLVM tools. One key feature of `pass:[libc++]` is that it supports deployment to older versions of macOS and iOS.
+
+. `pass:[libstdc++]` is the GNU Standard Library, which comes as the default with the GCC compiler. It is a robust and mature library, supporting many pass:[C++] standards and extensions. It's used by a large number of projects and is available on a wide variety of systems.
+
+You can specify which standard library you want to use by setting the appropriate flags in your build system. This is dependent on the compiler you're using.
+
+If you're using the Clang compiler and you want to use `pass:[libc++]`, you can use the `-stdlib=libc++` compiler flag. For example:
```
sdkrystian, line 65: Clang already uses `libc++` by default, so perhaps it would be better for the example to use `libstdc++`?

sdkrystian: Ah, I see there is an example of using `libstdc++` with clang below. Maybe this section in unnecessary then

user-guide/modules/ROOT/pages/advanced-builds.adoc
```diff
@@ -1,3 +1,247 @@
...
+./b2 toolset=clang cxxflags="-stdlib=libc++" linkflags="-stdlib=libc++"
+----
+
+
+
+== Specify Static or Dynamic Linking
+
+The compiler options `link=static` and `link=dynamic` specify how the Boost libraries should be linked to your project. If you do not specify a type of linking, `dynamic` is the default.
+
+. The `link=static` option specifies static linking. When you link statically, the linker takes the object code from the library and includes it in the executable file for the program. This leads to larger executables, but the resulting program is self-contained and does not require the library to be present on the system where the program is run. It can be beneficial when you want to avoid dependencies on shared libraries, particularly when you distribute your software to other systems.
```
sdkrystian, line 92: Extraneous period at the beginning of the line

user-guide/modules/ROOT/pages/advanced-builds.adoc
PeterTurcan: Build Scenarios?

user-guide/modules/ROOT/pages/advanced-builds.adoc
PeterTurcan: agreed, removed

user-guide/modules/ROOT/pages/advanced-builds.adoc
PeterTurcan: ok

user-guide/modules/ROOT/pages/advanced-builds.adoc
PeterTurcan: agreed, removed

user-guide/modules/ROOT/pages/advanced-builds.adoc
PeterTurcan: ok, removed this section

user-guide/modules/ROOT/pages/advanced-builds.adoc
```diff
@@ -1,3 +1,247 @@
...
+[source,txt]
+----
+clang++ -stdlib=libstdc++ your_file.cpp -lboost_system
+----
+
+When you're building the Boost libraries themselves, you can also specify the standard library to use with the `cxxflags` and `linkflags` options. The following command will build the Boost libraries using Clang with `pass:[libc++]`.
+
+[source,txt]
+----
+./b2 toolset=clang cxxflags="-stdlib=libc++" linkflags="-stdlib=libc++"
```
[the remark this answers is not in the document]
PeterTurcan: ok, updated

user-guide/modules/ROOT/pages/advanced-builds.adoc
PeterTurcan: ok, removed this section

user-guide/modules/ROOT/pages/advanced-builds.adoc
PeterTurcan: removed

user-guide/modules/ROOT/pages/advanced-builds.adoc
PeterTurcan: moved to the top

user-guide/modules/ROOT/pages/advanced-builds.adoc
PeterTurcan: removed all references to jamfiles

user-guide/modules/ROOT/pages/advanced-builds.adoc
PeterTurcan: ok

user-guide/modules/ROOT/pages/advanced-builds.adoc
PeterTurcan: removed it

user-guide/modules/ROOT/pages/advanced-builds.adoc
PeterTurcan: ok

user-guide/modules/ROOT/pages/advanced-builds.adoc
alandefreitas: Have a look at other options. I think this is `bash`.

See https://github.com/cppalliance/mrdox/blob/07471c21416f0332d89c5260729bf86a914ca240/docs/modules/ROOT/pages/usage.adoc?plain=1

PeterTurcan:
Expanded section on building individual libraries. Feedback included on the rest of the topics.

FYI: @alandefreitas @anarthal

alandefreitas (changes requested):

user-guide/modules/ROOT/pages/build-options.adoc
```diff
@@ -0,0 +1,152 @@
...
+Note:: Some libraries are header-only, meaning they do not need to be built before use. Examples include boost:asio[], boost:bind[], boost:function[], and boost:lambda[]. You don't need to use `--with-` for these libraries since they're included directly in your source code and don't need to be pre-compiled.
+
+To build multiple libraries, say boost:thread[] and boost:filesystem[], simply repeat the parameter:
+
+[source,bash]
+----
+b2 --with-thread --with-filesystem --with-system
+----
+
+The reason why we need `--with-system` is that it is a dependency of boost:filesystem[].
```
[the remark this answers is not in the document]
PeterTurcan: @alandefreitas I did not get this section right. If I build a boost library that can be built (ie: not header-only, such as filesystem) then indeed the dependencies are built too. The problem arises when I try to build a boost library like asio (which is header-only) BUT it has dependencies on libraries that need to be built (system etc.).  In this case the b2 command simply errors out - I have some mail out trying to verify how this should be handled.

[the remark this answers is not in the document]
alandefreitas: > The problem arises when I try to build a boost library like asio (which is header-only) BUT it has dependencies on libraries that need to be built

Yes. That makes more sense. I think the way `b2` looks at it (because it sucks in that regard) is that you can't build asio because it's not a "real" target in the `b2` sense because there's nothing to build so there can't be a `--with-asio` and then it can't find its dependencies because it doesn't exist. The way `cmake` handles it is much simpler because it sees `boost::asio` as an interface target that just propagates its properties, including any flags, include paths and dependencies.

If it's all about header-only libraries only, then the section looks OK as long as we adapt the example to a built library then to a header-only library to show this as a limitation. See the other point I also made about the `cmake` section though.

PeterTurcan:
@alandefreitas @anarthal @sdkrystian Thanks for the great feedback. I am going to close this PR and break the topics down into smaller chunks that I can wrap my mind around. I have saved off all the feedback and will be incorporating it into several PRs with narrower scope.

PR #124, opened 2023-07-28
boost build ==> b2.

grafikrobot:
The name is B2.

### DIFF
--- a/user-guide/modules/ROOT/pages/getting-started.adoc
+++ b/user-guide/modules/ROOT/pages/getting-started.adoc
@@ -294,7 +294,7 @@ This command should output the Boost version installed on your system.
 
 [tabs,sync-group-id=build]
 ====
-Boost.Build (Recommended)::
+B2 (Recommended)::
 +
 --
 . Navigate to https://www.boost.org/users/download/ and under *Current Release/Downloads*, download the Boost distribution.
@@ -304,7 +304,7 @@ For instance:
 
 . After the download has completed in your *Downloads* folder, right-click on the compressed file and extract it to a folder.
 
-TIP: boost:build[] is the officially supported build system used by the Boost libraries.
+TIP: b2[] is the officially supported build system used by the Boost libraries.
 --
 
 CMake::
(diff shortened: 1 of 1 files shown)


user-guide/modules/ROOT/pages/getting-started.adoc
```diff
@@ -304,7 +304,7 @@ For instance:
 
 . After the download has completed in your *Downloads* folder, right-click on the compressed file and extract it to a folder.
 
-TIP: boost:build[] is the officially supported build system used by the Boost libraries.
+TIP: b2[] is the officially supported build system used by the Boost libraries.
```
alandefreitas, line 307: In contexts with a `[]`, `boost:build[]` is a macro that is replaced with links to the documentation of tools and libraries in boostorg. 

image

`b2[]` is just a literal `"b2[]"`

image

alandefreitas:
`boost:build[]` is equivalent to `https://www.boost.org/tools/build[Boost.Build]`. It's somewhat analogous to `[@boost:/tools/build/index.html]` in quickbook (I might have got the qbk syntax wrong here).

What is the behavior you would like instead?

alandefreitas:
> Yes. Again, that's what I figured. But, again, where's the code that does that?

I linked the script in my first comment: https://github.com/cppalliance/site-docs/blob/develop/extensions/boost-link-inline-macro.js

> As is obvious from the PR.. Boost.Build => B2.

What to do about the text is obvious. I was asking what you want to do about the links we are going to replace:

1) The text should be B2. That's obvious.
2) Should the link still be `https://www.boost.org/tools/build`?
3) Do you want B2 to be implemented as an exception in the script above or just change these links? In other words, do you want this behavior for anyone who ever writes `boost::build[]` anywhere or do you just want the content writers of this repository to always use `https://...[B2]` instead of the macro?
3b) If you want this to be at the extension level, do you want it to be a replacement for `boost::build[]` or a new extra command, like `boost::b2[]`?

alandefreitas:
> replacing them with https://www.bfgroup.xyz/b2/[B2] would work.

okie dokie... `https://www.bfgroup.xyz/b2/[B2]` it is

PR #126, opened 2023-08-01
Table of Dependencies added

PeterTurcan:
@alandefreitas @anarthal @sdkrystian FYI


user-guide/modules/ROOT/pages/build-options.adoc
```diff
@@ -0,0 +1,4 @@
+= Build Options
```
alandefreitas, line 1: In the git history, it's best to have the file move/renamed and changed (see `git mv`) than to have a file removed and another file included.

user-guide/modules/ROOT/pages/library-dependencies.adoc
```diff
@@ -0,0 +1,1164 @@
+# Library Dependencies
+:navtitle: Library Dependencies
+
+The following table shows the _primary dependencies_ of the current Boost libraries. A primary dependency is a dependency explicitly referenced by the library headers. A _secondary dependency_ is any library that is referenced only by a primary dependency, and not by the library itself. The table also shows whether a library is _buildable_ or not.
```
alandefreitas, line 4: Was this information collected with boostdep and then post-processed with some script?

I usually don't like the idea of any information that depends on the current version of boost unless there's a clear low-effort (preferably zero-effort) strategy to maintain that information up to date once we move on to other projects, or we're going to end up with the same problem as the old documentation. In many cases, it might also add a lot of content with little information (in the sense that information is defined as the minimal representation of the data as less data and some process).

I see three solutions for these cases:

1) We include instructions on how users can generate this data themselves
2) We provide links to external resources that maintain this data, such as https://pdimov.github.io/boostdep-report/ and https://alandefreitas.github.io/boostdep_graph/
3) We include a python script that generates this adoc pages as partials and include the partials or an Antora JS extension script to render this data. In the case of python scripts, they could be run from time to time in CI. It could also be a single script that generates all partials.

Other people might have better ideas.

PeterTurcan: My 2c:
Currently the boostdep tool is not easy enough to use - it requires a modular version of boost which for most people will mean cloning the superproject, and following a series of command line instructions to set this up. The tool does produce good information, and this can be as an HTML table though is not as useful as the table I created as none of the dependencies are themselves links - so I cannot follow the path of dependencies - nor can additional columns such as "buildable" or "secondary dependencies" or any other information be added. Nor is there good navigation - I can easily find libraries that begin with "M" in my table, not so in the tools HTML output - there is no alphabetical index.
I am not sure of the value of asking every developer to go through this, when the information is identical for everyone and only needs to be done once, by us, and published in the most useful form. It is not an onerous task to reproduce this table for each release - an issue could be created for each release and filed on me. And each table should have the title of the version number, so users of older versions can locate the dependencies easily.

alandefreitas: > Currently the boostdep tool is not easy enough to use - it requires a modular version of boost which for most people will mean cloning the superproject, and following a series of command line instructions to set this up. The tool does produce good information, and this can be as an HTML table though is not as useful as the table I created as none of the dependencies are themselves links - so I cannot follow the path of dependencies - nor can additional columns such as "buildable" or "secondary dependencies" or any other information be added. Nor is there good navigation - I can easily find libraries that begin with "M" in my table, not so in the tools HTML output - there is no alphabetical index. I am not sure of the value of asking every developer to go through this, when the information is identical for everyone and only needs to be done once, by us, and published in the most useful form. 

Yes. I see the value of a better index. I don't think anyone would disagree with that side of the story. There's no value in asking every developer to go through this. Everyone agrees it's better to have better things. The discussion is about the process to keep this information up to date so we can have better things that are *also* reliable. Let me try to elaborate on that because I might have expressed myself poorly.

The point is about the other side of the story: the cost and risk of maintaining information that's only relative to the current version of boost, especially when it's one prone to change. Not having this process means all pages potentially contain information that's wrong and we're not doing anything to address it. There are lots of sites with old information about boost (for instance, https://theboostcpplibraries.com/ is one of the first when you google it) and this is hurting us right now because users use these resources to learn about boost and then complain about things that have been fixed long ago. We need a process where it's clear what/when/how to update things. By anyone. Also, for each such kind of information, some process to remind us or at least allow us to remember that the information needs to be updated.

Information that depends on the current state of the libraries is very special. Right now, we don't have any process to maintain that. For instance, there's one-page saying Boost.URL fuzz tests don't implement seeds. But it does now because there was an issue about it and it was fixed. So we already have pages from 1 month ago with information that's wrong. I could just go there and fix this one myself, but I just found out about it because of a tangential discussion I was having with Ruben somewhere else, or I wouldn't even know about it. On the other hand, someone else is even less likely to fix it, because they would need to know that this thing changed. We can wait for an issue, but this might never come and when it does the information was misleading users for a long time already. 

Of course, some information is more sensitive than others. No one really cares that much if Boost.URL uses seeds in fuzz tests. But some information is quite sensitive, like installation instructions, and the dependencies between libraries so the user can, for instance, fetch content from their cmake scripts. These can cause a lot of damage. Of course, the fix is much simpler in Boost.URL case: we simply don't write any sentences saying library X doesn't support Y, unless not supporting Y is part of the design rationale of X. Other statements should be fine. 

The general problem is there's no process to ensure this information is up to date and the expectations from each party should be clear whenever we include information like this. Whose responsibility is it to keep this up to date? Let's say person P is responsible for this. Is P responsible for maintaining a list of all information that might change and rechecking all repositories every number of days to ensure nothing has changed? Or will the information be wrong and mislead users until someone opens an issue? And what is the cost of updating this thing each time? A script might take 5 minutes. Doing it by hand could take hours. If it's hours, who's going to have these hours available?

Users don't tend to open issues when they find something that doesn't work unless they are already invested in the tool. If something is wrong, they won't even be sure if the information is wrong or if they are the ones doing something wrong. They'll usually just assume Boost sucks and move one. Only people invested in the thing are more likely to open issues. So we're going to see issues like Rene wanting it to be `B2` instead of `Boost.Build` more very few issues from random user saying some information was wrong. 

There's no process to keep this up to date, and the informal process seems to be that the information be wrong and mislead users until someone opens an issue. This is not good. Yes. I see the value of a better index. I don't think anyone would disagree with that side of the story. It's the other side of things we have to worry about.

> It is not an onerous task to reproduce this table for each release

It's hard to evaluate if this is onerous without looking at what the process is. 

Since there's no script I can only assume this was done by hand. It looks very onerous to me compared to running a script.

> - an issue could be created for each release and filed on me. 

Why do we need an issue if it's something we have to do in every release?

If there's no person allocated to this task, it won't be fixed with or without an issue.

The release managers are unlikely to take on this responsibility. Anyone else can't ensure they will have the time to do it.

> And each table should have the title of the version number, so users of older versions can locate the dependencies easily.

That's much more complex. This is a lot of information and these pages are not versioned. In any case, this is definitely the case where there's no way to keep this without a script.

So, it's in this context that I made that list of options:

- We don't provide the information because we don't want to pay this cost, which is inconvenient on the user and there's no point discussing the way that's a bad thing (We include instructions on how users can generate this data themselves)
- We provide information from external resources that is worse than it could be but can ensure the user the information is correct and fix the problem without the cost of maintaining it (We provide links to external resources that maintain this data, such as https://pdimov.github.io/boostdep-report/ and https://alandefreitas.github.io/boostdep_graph/)
- We do this properly with a script and have the information up to date (We include a python script that generates this adoc pages as partials and include the partials or an Antora JS extension script to render this data. In the case of python scripts, they could be run from time to time in CI. It could also be a single script that generates all partials.)

PeterTurcan: Agreed this should be fully scripted (after a round of feedback on navigation/format etc.) and versioned appropriately. To me, dependencies seem important enough that this is something we should take on, especially if we want to evangelize --with-<lib> builds. @anarthal @sdkrystian - any thoughts?

alandefreitas: > To me, dependencies seem important enough

Agreed

> that this is something we should take on, 

Yes. But being important doesn't necessarily mean *we* have to take on that because we also have to consider that other people already took on that, developed tools that are much more mature, and these resources are already available. So there's no point in coming up with something worse than what we already have.

For instance, a simple script could just take all information from https://pdimov.github.io/boostdep-report/, convert it to adoc using pandoc and put that in the partials. The information would be pretty, convenient *and* reliable. Or we could just use boostdep directly and format it in any way we want in the script, which is basically what https://alandefreitas.github.io/boostdep_graph/ does. Maybe Peter would even be willing to include an adoc generator to boostdep. Who knows.

> especially if we want to evangelize --with- builds. 

I should mention `--with-` builds but I don't really think we need to "evangelize" it. There are not so many compiled libraries in Boost and the user installing boost locally will most likely just want to build and install everything. Going back every time, building with `--with-one-more-lib` and reinstalling would take much more time. On the other hand, when we're in CI or using a package manager, we don't download all modules and then B2 will just build the ones we downloaded. Because of that, I think I never really needed `--with-` myself. 

That's not to say this information is not useful. There are lots of other cases where it's very useful to know the dependencies. I check that graph almost every day because of some decisions I have to make concerning library design. Users also often want to check if a library has too many unnecessary dependencies, which is usually a reason to avoid them. Just yesterday I had to check that graph to write a CMake script that would fetch a list of boost modules without going through boostdep since this script is using a fixed version of boost.

PeterTurcan:
Closed pending further discussion.

PR #131, opened 2023-08-15
Updated Headers, Separate Compilation topics

PeterTurcan:
Read and edited Headers and Separate Compilation legacy sections

### DIFF
--- a/contributor-guide/modules/ROOT/pages/design-guide/headers.adoc
+++ b/contributor-guide/modules/ROOT/pages/design-guide/headers.adoc
@@ -3,10 +3,7 @@
 :idprefix:
 :idseparator: -
 
-*Legacy content, currently being updated.*
-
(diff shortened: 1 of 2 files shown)


alandefreitas:
Looks good.

I've never seen the `BOOST\_FURBALL\_HPP` though. I didn't even know `\_` was valid C++.

It's worth going through the libraries to make sure that's the case, but I think all libraries I've seen so far follow a convention like `BOOST_<LIBRARY>_<PATH>_HPP`. For instance:

https://github.com/boostorg/url/blob/develop/include/boost/url/detail/any_params_iter.hpp

PeterTurcan:
Agreed on the header guards, clarified and added an example.

alandefreitas:
Great. Please squash when ready. :+1:

PeterTurcan:
@alandefreitas - squashed

PR #138, opened 2023-09-19
Added specific lib builds to Simulation scenario

PeterTurcan:
Added B2 and CMake process to build just the selected simulation libraries. Plan to add these commands for all other scenarios, it if makes sense to do so.

### DIFF
--- a/user-guide/modules/ROOT/pages/task-simulation.adoc
+++ b/user-guide/modules/ROOT/pages/task-simulation.adoc
@@ -19,3 +19,106 @@ Creating a real-time simulation of vehicles involves various aspects, including 
 * boost:mpi[] or boost:asio[]: For distributed simulations that run across multiple systems, you might need a library for network communication. boost:mpi[] provides a pass:[C++] interface for the Message Passing Interface (MPI) standard for distributed computing. boost:asio[] can also handle networking tasks and it is a bit lower-level.
 
 * boost:serialization[]: To save the state of the simulation or to communicate complex data structures over a network, you might find this library helpful.
+
+== Build only the Specific Libraries
+
+If you decided to run with the collection of libraries listed above, and only want to build these specific libraries, and their dependencies, then go through the following procedure:
+
+. Download the full Boost libraries package, and extract the files. Unless of course you have done so already. Refer to xref:getting-started.adoc[] for details.
+
+. Go through the bootstrap process. Again, unless you already have.
+
+. Optionally, create a folder for your specific build, say `C:\Users\YOURNAME\Documents\Boost\Simulation`.
+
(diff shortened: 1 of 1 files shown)


user-guide/modules/ROOT/pages/task-simulation.adoc
```diff
@@ -18,4 +18,107 @@ Creating a real-time simulation of vehicles involves various aspects, including
 
 * boost:mpi[] or boost:asio[]: For distributed simulations that run across multiple systems, you might need a library for network communication. boost:mpi[] provides a pass:[C++] interface for the Message Passing Interface (MPI) standard for distributed computing. boost:asio[] can also handle networking tasks and it is a bit lower-level.
 
-* boost:serialization[]: To save the state of the simulation or to communicate complex data structures over a network, you might find this library helpful.
\ No newline at end of file
+* boost:serialization[]: To save the state of the simulation or to communicate complex data structures over a network, you might find this library helpful.
+
+== Build only the Specific Libraries
```
alandefreitas, line 23: This looks like a huge section repeating the content of "Getting Started". Assuming it's a good idea for "scenarios" pages to repeat this content, why would it be only about `task-simulation.adoc` and not other "scenarios"?

user-guide/modules/ROOT/pages/task-simulation.adoc
```diff
@@ -18,4 +18,107 @@ Creating a real-time simulation of vehicles involves various aspects, including
...
+cmake .. -DCMAKE_INSTALL_PREFIX=C:\Users\YOURNAME\Documents\Boost\Simulation
+----
++
+This will generate build files in the build folder, and set the installation prefix to the specified folder.
+
+. After the build files are generated, use CMake to build and install:
++
+[source,cpp]
+----
+cmake --build . --target install
```
alandefreitas, line 115: This is the command for installing, not building (although cmake might notice a dependency between two targets).

The cmake script above is also not prepared to install the library. Creating a install target is quite complex and even some good libraries don't provide these. 

A user experimenting with boost will definitely not create a `install` target. They will basically build and run code. Installing it as an application is a complex issue related to cmake only.

Have you run any of these commands and scripts locally? It's hard to believe any step worked at all.

alandefreitas (commented):
Where is this page located in the navigation bar? Shouldn't the link be `user-guide/modules/ROOT/pages/.../task-simulation.adoc` instead of `user-guide/modules/ROOT/pages/task-simulation.adoc`?

PeterTurcan:
This section does not seem to add much value, as most boost libraries are required as dependencies - seems to make most sense to install all of them.

PR #139, opened 2023-09-22
Initial migration of Users/People pages in adoc format

4down:
Porting pages that are linked on the new website over to adoc format in order to minimize or eliminate any links taking a user offsite.


alandefreitas:
This looks very weird. It's not even in any of the documentation components. Where did this requirement come from?

4down:
@alandefreitas This is the start of porting pages such as https://www.boost.org/users/people.html and all linked pages off of it so that they can live on the new site.  Currently we have a few pages linking to boost.org on the new site, and it would be nicer to not have those be offsite links.  

Anything under /site-pages/ are static content pages for the website that we want people to still be able to contribute to editing.

alandefreitas:
Oh... OK. Did this `site-pages/users/images/` directory layout come from the old website? This looks weird because these people are not users. Some have even passed away already.

4down:
@alandefreitas https://www.boost.org/users/people.html

alandefreitas:
> https://www.boost.org/users/people.html

I would remove or replace `users` from this url but it's up to you. We can always change it later as long as the URL in the website is stable.

Issue #149, opened 2023-11-08
Provide "Modular Boost" section


akrzemi1:
Apart from the old Boost website, Boost used to have a Boost Trac Wiki with a lot of useful information. One of them was "how to start with modular Boost": https://svn.boost.org/trac/boost/wiki/StartModMaint.

This is a request to resurrect that page as part of the new docs.

PeterTurcan:
I can look into porting the pages to the new site.

alandefreitas:
I don't have access to this content but we did include information about modularization in the new getting started section. We talk about boostdep/depinst and vcpkg.

akrzemi1:
Try this one: http://trac.cpp.al/trac/boost/wiki/StartModMaint
It is titled "Getting Started with Modular Boost Library Maintenance" and it tells how to use git for Boost maintainers.

alandefreitas:
Oh... Nice. It's intended for contributors. The getting started guide is for users. Great content.

PeterTurcan:
I assume the task would be to migrate the  "Getting Started with Modular
Boost Library Maintenance" page, and all of the pages it links to. A few of
the links do not appear to go anywhere specific or useful - so will need
updated, though five pages that seem valid are:

*http://trac.cpp.al/trac10/wiki/TortoiseGitLibMaint
<http://trac.cpp.al/trac10/wiki/TortoiseGitLibMaint>*

*http://trac.cpp.al/trac10/wiki/ModBigPicture
<http://trac.cpp.al/trac10/wiki/ModBigPicture>*

*http://trac.cpp.al/trac10/wiki/StartModWorkflow
<http://trac.cpp.al/trac10/wiki/StartModWorkflow>*

*http://trac.cpp.al/trac10/wiki/StartGitHub
<http://trac.cpp.al/trac10/wiki/StartGitHub>*

*http://trac.cpp.al/trac10/wiki/PostCvtMergePoint
<http://trac.cpp.al/trac10/wiki/PostCvtMergePoint>*

These five pages link to further items, but not other wiki pages that would
need to be converted - so the task looks like converting these 6 pages to
ADOC files, and updating some links that are no longer available or
useful.  Perhaps we need a new top level node in the Contributor Guide,
such as "Working with Modular Boost" - or similar.

Does that make sense?

- Peter

PR #154, opened 2023-11-19
Add Modular Boost section - six pages

PeterTurcan:
Added Modular Boost sections - from wiki pages and boostdep steps to install it, and related images. Added to the Contributor Guide, though perhaps it belongs in the User Guide?


contributor-guide/modules/ROOT/nav.adoc
```diff
@@ -33,6 +33,14 @@
 ** xref:testing/continuous-integration.adoc[]
 ** xref:testing/fuzzing.adoc[]
 
+* Modular Boost
```
alandefreitas, line 36: This content was originally written a long time ago and I think this concept of "Modular Boost" has changed over time. At the time, we had SVN boost in a single repository, and then boost git repositories came up. Because each library had its own repository, this was considered modular. Note that this is part of the contributor guide and didn't affect who the users consumed boost.

Nowadays, when people talk about modularizing boost, they are talking about the ability of the _user_ to fetch only the boost modules they need as individual libraries in one way or another. That's why the new user guide includes content about how to download only individual modules with their dependencies and how to consume individual libraries as packages.

This distinction would make things a little confusing for people reading it nowadays. What we used to call "modular" boost at that time, is not enough for people to consider boost modular. The old term is better represented by the "Boost repositories layout" or something like that. 

We should get feedback from @akrzemi1 in this PR.

The commit message should also include `fix #149` in its body.

contributor-guide/modules/ROOT/pages/modular-boost/establish-merge-point.adoc
```diff
@@ -0,0 +1,81 @@
+= Establishing a merge point after SVN Conversion
```
alandefreitas, line 1: I'm not sure this is relevant nowadays because this transition is already complete for any new contributors. I think we should at least have an admonition explaining this transition is already complete.

contributor-guide/modules/ROOT/pages/modular-boost/library-workflow-overview.adoc
```diff
@@ -0,0 +1,72 @@
+= Modular Boost Library Workflow Overview
```
alandefreitas, line 1: Maybe we don't need the "Overview" suffix on all of these pages?

contributor-guide/modules/ROOT/pages/modular-boost/tortoisegit-library-maintenance.adoc
```diff
@@ -0,0 +1,27 @@
+= TortoiseGit Library Maintenance
```
alandefreitas, line 1: We could confirm that but I don't think separate TortoiseGit is relevant at all anymore. Times have changed and people will either use the command line or whatever comes with their IDE.

PeterTurcan: OK, removed this section.

contributor-guide/modules/ROOT/pages/modular-boost/library-workflow-overview.adoc
PeterTurcan: Agreed, removed from title, nav-title and filename

contributor-guide/modules/ROOT/pages/modular-boost/establish-merge-point.adoc
PeterTurcan: If it is not needed, the section should be removed.

alandefreitas: Sounds good to me

contributor-guide/modules/ROOT/nav.adoc
PeterTurcan: @alandefreitas I have renamed the section "The Git Super-project" - which seems to be what it's about - and have reworked or removed all references to "modular boost" so we can re-use this term when needed.

alandefreitas: Oh! That's a great name. The super-project is what people are usually trying to understand. I think I would just call it "The Super-Project" because it's more powerful and less is more.

PeterTurcan: @alandefreitas OK, sounds fine, made that change and corrected all titles so that they are title-case, and squashed.

contributor-guide/modules/ROOT/pages/modular-boost/establish-merge-point.adoc
alandefreitas: The section is still here

contributor-guide/modules/ROOT/pages/modular-boost/library-workflow-overview.adoc
alandefreitas: Still here

contributor-guide/modules/ROOT/pages/modular-boost/tortoisegit-library-maintenance.adoc
alandefreitas: Section is still here. And also the images are still in the commit.

contributor-guide/modules/ROOT/pages/modular-boost/library-maintenance.adoc
```diff
@@ -0,0 +1,281 @@
+= Getting Started with Modular Boost Library Maintenance
+:navtitle: Library Maintenance
+
+This page briefly sketches the mechanics of maintaining a Boost library using Git and Modular Boost. The intended audience is developers getting started with Git and Modular Boost maintenance of existing Boost libraries.
+
+This page is intended to get you started only; it does not provide in-depth coverage. See links below for that.
+
+Illustrations of how to perform various actions are given using the Git command line client. Illustrations of the same actions using TortoiseGit are available on a separate xref:modular-boost/tortoisegit-library-maintenance.adoc[TortoiseGit Library Maintenance page].
```
alandefreitas, line 8: Still references to TortoiseGit

alandefreitas (commented):
It seems like the changes are still local

PeterTurcan:
Does the last commit remove the unwanted files?

alandefreitas:
> Does the last commit remove the unwanted files?

The images are still there: https://github.com/cppalliance/site-docs/pull/154/files

PeterTurcan:
Good catch - images deleted.

PR #155, opened 2023-11-27
Populating the Help page from the FAQ

### LINKED ISSUE #79: FAQ
There is a list of questions people have about Boost. Often these questions take the form of criticisms, because people don't understand the reasons for why Boost maintainers do things one way and not the other. There should be a FAQ page where these questions are answered and criticisms are addressed.

PeterTurcan:
Moving the user guide FAQ to the site-docs Help page. Doesn't seem like there should be conflicting sources, so will remove the FAQ from the User Guide in the next PR.

fix #79

### DIFF
--- a/site-pages/help.adoc
+++ b/site-pages/help.adoc
@@ -1,5 +1,361 @@
 = Help
 
-This is the help page.
-
 You can submit content for this page here: https://github.com/cppalliance/site-docs.
+
+== Frequently Asked Questions
+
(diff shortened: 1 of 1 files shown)


alandefreitas:
If we're just moving the content, the repo history is better if the file is just moved in a single PR. See `git mv`... The file from the antora docs can be removed from nav.adoc or you can create a symlink in `partials` so the content is always in sync.

PR #158, opened 2023-12-12
Add common dependencies to best dev practices

PeterTurcan:
Added a table of the 20 libraries most commonly depended on by other libraries, to the Best Practices section of the Contributor Guide.

### DIFF
--- a/contributor-guide/modules/ROOT/pages/best-practices.adoc
+++ b/contributor-guide/modules/ROOT/pages/best-practices.adoc
@@ -1,4 +1,39 @@
 = Best Practices
 :navtitle: Best Practices
 
-*Content under construction*
+This section contains guidance on how to reduce development time from concept through to publishing of a new library.
+
+== Common Dependencies
+
+Some libraries are published with the intended primary audience, or in some cases the sole audience, being developers of other libraries. These libraries are published to make some of the time consuming and awkward processes of Boost-compliance easier. It is good practice for a new library developer to read the introductions to each of these libraries, and ascertain if they might be of value to the library they are developing. The libraries in the following table are listed in the order of the number of other libraries that have included them as a primary dependency. The top eight listed below are a primary dependency of 50 or more other libraries.
+
+[cols="1,2,4",options="header",stripes=even,frame=none]
+|===
+| *Rank*  | *Library* | *Description* 
+| 1 | boost:config[] | Helps Boost library developers adapt to compiler idiosyncrasies
+| 2 | boost:core[] | A collection of simple core utilities with minimal dependencies
+| 3 | boost:type-traits[] | Templates for fundamental properties of types
+| 4 | boost:assert[] | Customizable assert macros
+| 5 | boost:static-assert[] | Static compile time assertions
+| 6 | boost:throwException[] | A common infrastructure for throwing exceptions from Boost libraries
+| 7 | boost:mpl[] | A general-purpose, high-level template metaprogramming framework of compile-time algorithms, sequences and metafunctions
+| 8 | boost:preprocessor[] | Preprocessor metaprogramming tools including repetition and recursion
+| 9 | boost:utility[] | A collection of general-purpose components for language support
+| 10 | boost:iterator[] | A system of concepts which extend the pass:[C++] standard iterator requirements and a framework of components for building iterators based on these extended concepts
+| 11 | boost:smart_ptr[] | Smart pointer class templates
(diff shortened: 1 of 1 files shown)


contributor-guide/modules/ROOT/pages/best-practices.adoc
```diff
@@ -1,4 +1,39 @@
 = Best Practices
 :navtitle: Best Practices
 
-*Content under construction*
+This section contains guidance on how to reduce development time from concept through to publishing of a new library.
+
+== Common Dependencies
```
alandefreitas, line 6: Maybe "usual dependencies"? By "common dependencies" I thought it would be about dependencies common to all libraries.

contributor-guide/modules/ROOT/pages/best-practices.adoc
```diff
@@ -1,4 +1,39 @@
...
+
+Some libraries are published with the intended primary audience, or in some cases the sole audience, being developers of other libraries. These libraries are published to make some of the time consuming and awkward processes of Boost-compliance easier. It is good practice for a new library developer to read the introductions to each of these libraries, and ascertain if they might be of value to the library they are developing. The libraries in the following table are listed in the order of the number of other libraries that have included them as a primary dependency. The top eight listed below are a primary dependency of 50 or more other libraries.
+
+[cols="1,2,4",options="header",stripes=even,frame=none]
+|===
+| *Rank*  | *Library* | *Description* 
+| 1 | boost:config[] | Helps Boost library developers adapt to compiler idiosyncrasies
```
alandefreitas, line 13: How did we come up with this list? Some are even reasonably high-level:

https://alandefreitas.github.io/boostdep_graph/

I would only think about config, core, asset, static-assert, and throw-exception. For instance, you can see how almost all libraries depend on config:

https://alandefreitas.github.io/boostdep_graph/libs/config.html

All other libraries are libraries I don't see being used that much as dependencies. Some have even become obsolete by the standard libraries or the language itself so they are not recommended in new contributor projects. For instance, `std::move` replaced `boost::move` since C++11 and the minimum standard for boost is now C++11.

These libraries (or most of their features) are already supported by C++11 and people just use C++ features directly:

- boost:type-traits
- boost:smart_ptr
- boost:function
- boost:move
- boost:array
- boost:bind
- boost:typeof

And these are obsoleted by other boost libraries or overlap in functionality so that people usually don't go to them first:

- boost:mpl: people usually recommend boost:mp11 nowadays
- boost:utility: people usually start with boost:core and just include boost:utility if they really need it for some reason (they are both collections of utilities)
- boost:predef: people usually start with boost:config and just include boost:predef if they really need it for some reason (they are both collections of macros)
- boost:preprocessor: people usually just use boost:config

These are not really needed in most libraries and some are even high-level libraries:

- boost:iterator
- boost:integer
- boost:range (also obsoleted by C++20 ranges)
- boost:optional (also included in C++17)
- boost:container-hash (also very specific to the library type)
- boost:concept-check (also obsoleted by C++20 and SFINAE is usually used as a replacement before that)

alandefreitas (changes requested):

contributor-guide/modules/ROOT/pages/best-practices.adoc
PeterTurcan: @alandefreitas  FYI: the list comes from running a script over the primary dependencies listed by boostdep, and simply counting the number of times each lib appears as a primary dependency. The counts were:
config		132
core		        102
type_traits         96
assert		85
static_assert	75
throw_exception	65
mpl		        55
preprocessor	50
utility		41
iterator		39
smart_ptr		38
function		25
integer		25
predef		24
range		23
move		23
optional		23
container_hash	21
array		20
bind		20
concept_check	20
typeof		18

Agreed that dated libraries or those now covered in the standard should be dropped from any best practices.

alandefreitas: I see. These are direct dependencies. Except for type_traits, your top 6 is basically the list I suggested. type_traits used to be very useful but doesn't make sense anymore because new libraries can get the same functionality from C++ directly now. And anything after throw_exception is not so common and doesn't make much sense in new libraries.

PeterTurcan: OK, reduced the list to the top 6, and added mp11 for metaprogramming, and changed the name to Beneficial Dependencies.

PR #179, opened 2024-03-23
Add security section to contributors faq

PeterTurcan:
Added nine questions/answers particular to Security from a contributor's perspective.


alandefreitas:
Nice. Where did this content come from? Is it good to go?

PeterTurcan:
Thanks for the feedback, yes GTG - I came up the questions based on what I would ask if I was thinking about writing a library - and then researched using internet tools.

PR #198, opened 2024-04-02
Stack Overflow call updated

### LINKED ISSUE #189: Update StackOverflow link in User Guide/Resources
https://www.preview.boost.org/doc/user-guide/resources.html#_stack_overflow

The link text is "stack overflow: Boost Library", where "overflow" is formatted bold. The website is typically called StackOverflow (in one word or sometimes two, separated with a space). I think, the text should be updated accordingly.

The link points to a search query with an "s=d2d13545-4fbc-4f13-a029-08e8a70d2bdc" parameter. I'm not sure what this parameter means, but I suspect it is related to a user session. I think, this parameter should be removed.

Furthermore, it may be better to use the "boost" tag instead of the search query, i.e.:

https://stackoverflow.com/questions/tagged/boost

PeterTurcan:
Updated the call into Stack Overflow:

fix #189

Stack Overflow should be capitalized, as like Boost its logo is lower case but all other references should be caps.


alandefreitas:
Great. Just reminding you don't need a new PR for this kind of thing. Just update the branch associated with the PR (`PeterTurcan:develop` in this case) by force-pushing to it and the PR will be updated.

PR #203, opened 2024-04-08
Add Writing a Review topic

PeterTurcan:
Populated the Writing a Review topic of the Formal Reviews guide, and added links to and from the Contributor Guide.  A partial fix to issue 181.


alandefreitas (commented):
There's a template for these reviews. It's be good to include that.

I don't know where it is in the old website, but the review manager usually posts the template whenever there's a new review.

PeterTurcan:
Good idea. Added a template based on email submissions of recent libraries, and updated that template with specific questions that should be helpful to a reviewer.

PR #220, opened 2024-04-24
Update vcpkg integration information

Osyotr:
People should not install all boost libraries by default.

### DIFF
--- a/user-guide/modules/ROOT/pages/getting-started.adoc
+++ b/user-guide/modules/ROOT/pages/getting-started.adoc
@@ -121,50 +121,27 @@ They provide a simple and consistent way to download, configure, and install lib
 Vcpkg::
 +
 --
-To install Boost with Vcpkg, you can run a command like:
-
-[source]
-----
-vcpkg install boost
-----
-
-or adding `boost` to your `vcpkg.json` manifest file:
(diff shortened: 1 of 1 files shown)


user-guide/modules/ROOT/pages/getting-started.adoc
```diff
@@ -121,50 +121,27 @@ They provide a simple and consistent way to download, configure, and install lib
 Vcpkg::
 +
 --
-To install Boost with Vcpkg, you can run a command like:
```
alandefreitas, line 124: The classic/manifest mode issue is independent from "People should not install all boost libraries by default.".

PeterTurcan:
Personally I think this update should not be merged. Installing individual Boost libraries is not as simple as it sounds given the complexity of primary and secondary dependencies. For newcomers to Boost installing all libs is the simplest way to go that expedites getting something to work (this is the "getting started" section after all).
Possibly a note could be added that installing everything is not the most efficient for experienced devs, but I would not encourage the installation of individual modules as a first step. I tried it by installing only the six or so libraries mentioned in the Simulation section, only to find that 130+ libs get installed given the tree of dependencies.
Open to correction if I am not fully understanding the issue.

PR #268, opened 2024-07-29
Standard Library section added to User FAQ

PeterTurcan:
Starting a new section in the User Guide FAQ to cover questions about the relationship between Boost libraries and the Standard Library.

### DIFF
--- a/user-guide/modules/ROOT/pages/faq.adoc
+++ b/user-guide/modules/ROOT/pages/faq.adoc
@@ -675,6 +675,18 @@ void foo() {
 }
 ----
 
+== Standard Library
+
+. *How can I be sure when I should use a Boost library or a component of the Standard Library?*
+
+Some Boost libraries provide useful and advanced functionality unavailable in the Standard Library. Other Boost libraries have indeed been superseded by the Standard Library, but remain in Boost for backwards compatibility. To determine which you should use, given the choice, consider working through the following process.
+
+.. Check the Boost library documentation, as their relationship to the Standard Library is often documented. Both the overview and the release notes are good sources of information for mentions of standardization.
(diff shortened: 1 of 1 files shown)


alandefreitas (commented):
> Some Boost libraries provide proper and advanced functionality unavailable in the Standard Library. Other Boost libraries have indeed been superseded by the Standard Library, but remain in Boost for backwards compatibility.

This description is not very accurate. 

Many Boost libraries became part of the standard, but not most. Most libraries are only available in Boost. And for the ones that did, usually only a subset of the libraries went into the standard.

So...

> Some Boost libraries provide useful and advanced functionality unavailable in the Standard Library

Actually, _most_ Boost libraries provide functionality unavailable in the Standard Library. The few that went into the standard also usually provided extra functionality that wasn't standardized.

> Other Boost libraries have indeed been superseded by the Standard Library, but remain in Boost for backwards compatibility

This is a complex issue because they often have extra functionality that is not standardized. 

I'd say we have three categories:

- The library has no relationship with the standard (For instance, Boost.Beast)
- The bulk of the library was standardized, but it offers extra functionality that wasn't (For instance, Boost.System)
- The whole library has been standardized, but it's still worth using the Boost version because of performance (For instance, Boost.Regex)
- The whole library has been standardized, and the only reason to use it would be backwards compatibility (For instance, Boost.Thread)

We have to identify which library is in each category. Most libraries would be in the first category. Unfortunately, the library documentation doesn't usually include a comparison with what's been standardized.

PeterTurcan:
@alandefreitas Good feedback, incorporated it into the answers, and added a few simpler questions.

alandefreitas:
That looks much better. Thanks!

PR #272, opened 2024-08-02
Adds type specimen

julioest:


rbbeeston (approved):
LGTM, but did we get all the tags we would need?

PR #294, opened 2024-08-12
Debug Visualisers section added to CG

### LINKED ISSUE #278: CG: Debug Visualizers section
Contributor's Guide needs a section on Debug Visualizers explaining how to use them and write them for MSVC and gdb.

PeterTurcan:
fix #278

This section could be added to the User Guide - as it really applies to all Boost developers.


alandefreitas:
> This section could be added to the User Guide, as it applies to all Boost developers.

If it applies to Boost developers, it's already where it should be. Developers are contributors. 

Users might want to know how to *use* the visualizer instead of how to develop a visualizer. But that's a different kind of content. However, we can't write that content yet because there's no common convention for using Boost visualizers.

PR #309, opened 2024-08-22
Couple Other Languages questions added to FAQ

PeterTurcan:


alandefreitas:
The content looks good, but is it related to specific issues? Most PRs I see have no issue attached to them. Is it because people are requesting content on Slack? If so, is there a way to track this?

rbbeeston:
@alandefreitas  Vinnie has a lot of quick daily changes he wants to see get into place. Due to the frequency and the number of tasks, Julio and I have been tasked with knocking them out, and just verify that they pass the checks for the time being. This was also discussed with Joaquin, before he went on vacation. This is a temporary and targeted situation related to a number of UI changes.

PR #310, opened 2024-08-26
Overflow handling for tables, pre, code container elements

julioest:


rbbeeston (commented):
in the cases I saw, I didn't see any issues with pre and code blocks. this change may cause problems with existing ones.

PR #331, opened 2024-09-13
update raw boostlook.css url path

julioest:


vinniefalco:
This needs to be "master" or "develop" depending on the branch being built.

vinniefalco:
We should still merge this, even without the fix for the branch, since it is better than what we have.

PR #363, opened 2024-11-18
Steering Committee topic added to CG

### LINKED ISSUE #326: CG: Explain Steering Committee
On Sun, Sep 8, 2024 at 8:42 AM Klemens Morgenstern via Boost <boost@lists.boost.org> wrote:
> It is also quite possible that new steering committee will actually be
> able to steer boost, if it is made up by developers. This would overcome
> issues in boost, because instead of commanding from on-high to migrate
> to CMake, the developers would first put in the work to build consensus
> (by building the thing) before announcing move. It looks to me like this has
> succeeded with CMake thanks to Peter and is in the works for a modularization
> by Rene.

Firstly thank you for your review, and apologies for pressing Ctrl+ENTER by accident.

It's important to understand that this new Steering Committee, despite the unfortunate name, is not responsible for "steering Boost:"

"The purpose of the Steering Committee is to make decisions regarding the use of the Boost assets which require ownership by an individual or legal entity."

This is important enough that we should probably add it to the Contributor's Guide (Turcan?) under "Steering Committee" so there is no confusion.

PeterTurcan:
fix #326


PeterTurcan:
@grafikrobot OK, thanks - will rewrite with the correct name - but does this mean this group is out of date, or needs its name changed too perhaps?
https://groups.google.com/g/boost-steering?pli=1

PR #387, opened 2025-01-06
add Unordered to libs playbook

cmazakas:
Depends on https://github.com/boostorg/unordered/pull/299

I'm just opening this now so we can pull the trigger later


alandefreitas:
```
[20:00:17.631] FATAL (antora): antora.yml not found in https://github.com/boostorg/unordered (branch: develop | start path: doc)
```

PR #410, opened 2025-02-20
update the old Boost logo #1475

cdw9:
https://github.com/boostorg/website-v2/issues/1475

I'm not sure that this copy of the logo is actually used anywhere. But it has the same dimensions as the old version, so it should not hurt anything to update it


rbbeeston (approved):
Looks good to me, go ahead and merge!

PR #431, opened 2025-03-28
Bumped CMake version to 3.5

DockedFerret800:
With the release of CMake 4.0, compatibility with versions older than 3.5 has been removed. Calls to cmake_minimum_required with older version result in a fatal error.


nigels-com:
Yes, agreed.

PR #461, opened 2025-06-05
update tweeting guide with high-res Boost logo

julioest:
There were blue artifacts in the previous image. I’ve updated the logo to a current high-resolution version.


rbbeeston (approved):
thanks! LGTM!

karimarie67:
Agree with Rob, it looks much better!

joaquintides:
Allow me chime in, while you're at it you may want to select an image without hashtags, and remove the section on hashtags from the tweeting guide. Hashtags are long frowned upon in X.

rbbeeston:
I mean, does frowned upon mean not allowed or "makes you look like a millennial who wonders why they can't get a job in social media"?

joaquintides:
> I mean, does frowned upon mean not allowed or "makes you look like a millennial who wonders why they can't get a job in social media"?

More like the latter: https://x.com/elonmusk/status/1869070358210572306

> But hashtags are still valid and useful in the other social channels. Which we populate from the X post.

I guess these can be added to the posts in those other social channels. FWIW, the last N posts to X fom Boost_libraries didn't have hashtags.

PR #464, opened 2025-06-20
Revised Logo Policy usage guide

### LINKED ISSUE #263: Logo Usage Policy needs work
The language of the logo usage policy needs work. The policy governing use of the logo in the documentation which goes into a release should be adjusted to ensure that third parties do not lose the ability to freely create derivative works which incorporate parts or all of Boost. How the logo is handled in this case needs to be well-defined.

And, this issue is appropriate for discussion of any areas of the policy, with the goal to ensure that a balance is struck between honoring the spirit of the BSL while protecting the logo from misuse.

PeterTurcan:
First cut at trying to update the Logo Policy to match your input - @vinniefalco, @grafikrobot 

fix #263

### DIFF
--- a/contributor-guide/modules/ROOT/pages/docs/logo-policy-media-guide.adoc
+++ b/contributor-guide/modules/ROOT/pages/docs/logo-policy-media-guide.adoc
@@ -9,54 +9,66 @@ Official repository: https://github.com/boostorg/website-v2-docs
 = Logo Policy and Media Guide
 :navtitle: Logo Policy and Media Guide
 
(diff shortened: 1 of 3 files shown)


PeterTurcan:
Closing this for now, pending input and consideration from the FSC.

PeterTurcan:
@grafikrobot Thanks for your feedback. Vinnie was not sure about the right thing to do here, given it was a document prepared by an IP lawyer, so we decided to shelve the update until we have a clearer understanding of the issues involved.

PR #504, opened 2025-09-15
Diagnostics topic added to User Guide/Development

PeterTurcan:
Created a Diagnostics topic, focusing on the symbols that seem to me to add significant value over the Standard. Small code samples demonstrate these symbols, and the code has been run and tested against Boost 1.88.0.
@vinniefalco - a first cut - too technical? about right for tone?
@pdimov - are there stronger use cases for BOOST_VERIFY, or boost::throw_exception that I could address in the short samples?

### DIFF
--- a/user-guide/modules/ROOT/nav.adoc
+++ b/user-guide/modules/ROOT/nav.adoc
@@ -38,6 +38,7 @@ Official repository: https://github.com/boostorg/website-v2-docs
 ** xref:counted-body.adoc[]
 ** xref:implementation-variations.adoc[]
 ** xref:reduce-dependencies.adoc[]
+** xref:diagnostics.adoc[]
 
 * User Community
 ** xref:user-community-introduction.adoc[]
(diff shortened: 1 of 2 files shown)


vinniefalco:
something is off. the information is not _wrong_ per se but it is not organized in a way that catches the eye. and the exposition is not generating within me positive emotions and excitement.

PeterTurcan:
@vinniefalco OK - I'll rework with more excitement in mind

vinniefalco:
The document should answer these questions:

"What are the most useful diagnostic facilities in Boost?"
"How can these facilities help me be more effective at my job?"

pdimov:
Some other comments:

*  `assert` is a macro, not a function, and there's no such thing as `std::assert`.
* `boost::system::error_code` does provide extra functionality that `std::error_code` lacks. It allows attaching source locations to error codes.

PeterTurcan:
@vinniefalco second cut - did use AI for some marketing style language that does not come naturally to me. Same code examples as before and focusing on the four calls I think add most value. 
@pdimov thanks for the feedback - always appreciated

PR #528, opened 2025-10-31
Add changes for Parser and STLInterfaces.

tzlaine:


PeterTurcan (changes requested):
Can you verify that the first three bullet points for the Parser library are the three most impactful updates (from the user's perspective) to this library? Maybe they are, just need to validate before merging, but if not re-order so the top three cover the most salient updates.

PR #530, opened 2025-11-03
Boost.Redis 1.90 release notes

anarthal:


anarthal:
This is release notes, not regular docs. I can coalesce things into sub bullet points if you prefer, but I can't change what the library has changed in the last release. I won't remove things from the list. And yes, they are already ordered in order of importance.

PR #544, opened 2025-11-19
Update release notes for Boost.Locale in 1.90.0

Flamefire:


PeterTurcan (changes requested):
Minor tense tweaks to be consistent with other release notes:

** Fixed B2 build files to avoid building dynamic versions of various Boost libraries
   when only static ones are requested (boost_gh:pr[locale,266]).
** When ISO8859-1 or ISO8859-8 encoding is requested, allow using Windows codepages 1252 or 1255, respectively,
instead of using the "C" (classic) locale when the selected Windows locale doesn't support that ISO8859 encoding.
** Enabled a workaround for an issue in Cygwins stdlib when converting some long UTF-8 sequences to UTF-16.

Flamefire:
Makes sense, thanks!

BTW: If you make the suggestion inline as a proposed diff next time, I can accept it directly which shows you as author if you wanted that.

The 2nd bullet point didn't had a change, did it?

PR #545, opened 2025-11-19
Reduce FOUT with font preloading

julioest:
### Problem: 
Fonts load late, causing a flash of unstyled text (FOUT) on initial page load.
### Solution: 
Add `head-fonts.hbs` partial with Font Loading API, preload fonts, and 3-second timeout fallback to ensure fonts load before content is visible.
### Changes:
- New `head-fonts.hbs` partial with preload links, Font Loading API implementation, and timeout logic

### Impact: 
Eliminates FOUT, fonts load earlier, graceful fallback for older browsers, better perceived performance.


sdarwin:
- notosans.woff2 continues to 404.  Twice.
- site.webmanifest 404's once.
- This solution is creating a dependency on the cppalliance.org website.  That's alright in the short term but longer term the doc build process, using antora, should generate standalone, self-sufficient doc builds that embed the required files, and aren't contacting cppalliance.org.     If possible.

Are all these "notosans" fonts the primary main fonts?    If they were 404ing before, does that mean they weren't displayed?   In that case, are they needed?

PR #561, opened 2025-12-11
Added stub for 1.91 release notes

joaquintides:
added stub for 1.91 release notes


sdarwin:
In this same PR, why not remove "Month Day, Year Hour::Minute GMT" from template.adoc?   The date is shown earlier on the webpage using Django.

PR #586, opened 2026-02-01
Updated release notes for LEAF and Exception

zajo:


PeterTurcan (changes requested):
Can this statement be edited to make it clear what was added, what improved since the last release?
Customizable serialization API, built-in support for Boost.JSON and nlohmann/json.

PR #607, opened 2026-03-03
Update release notes for Boost 1.91.0 for opitonal

akrzemi1:
Updated release notes for Boost 1.91.0, detailing breaking changes, enhancements, and new features in the optional and unordered libraries.


PeterTurcan (changes requested):
Like the detail, a few nits:

Correct the odd punctuation (single quote) of:     (are ['core constant expressions]) 

Remove "In the said implementation," and just start the three sentences with Abandoned... Applied... and Added....

In your Warning you say "may affect the overload..." then "the following code will behave differently" - seems like a conflict - both should be "may" if you are not yet certain, or "will" if you are certain that behavior will change. Good to warn developers of possible future behavior changes.

PR #608, opened 2026-03-10
Update boost_1_91_0.adoc

apolukhin:


apolukhin:
ping

PeterTurcan (changes requested):
Few nits:

1. Let's keep the tense all past tense, change these to 'ed':

Fixed 'import std' ....
Disabled `from_exception` ...

2. Clunky double-negative sentences, suggest changing:

 The library now does not check at compile time for a possibly incompatible runtime. As a result  the check now does not have false positives and the `from_exception` functionality  just works out-of-the-box.

to:

Removed the compile time check for a possibly incompatible runtime. As a result there are no false positives and the `from_exception` functionality  just works out-of-the-box.

And:

Simplified `has_addr2line` check to not link with C++ Standard Library to simplify builds with non system-default libstdc++.so

to:

Removed the `has_addr2line` link check with C++ Standard Library to simplify builds with non system-default libstdc++.so

3. Awkward sentence:

Change:   Size of binaries that use the library become smaller. 

to

Reduced the size of library binaries.

apolukhin:
@PeterTurcan many thanks for the review! Applied the proposed fixes

PR #610, opened 2026-03-17
Add 1.91.0 release notes for JSON

grisumbras:


PeterTurcan (changes requested):
Are you able to add a bit more context to this statement:

Other fixes and improvements.

Such as:
Other non-breaking improvements to the performance and documentation? 

Or similar?

grisumbras:
Sure, I can. But my usual approach is to only mention new features and major changes in Boost release notes, and mention less important changes in Json release notes. If you don't like "other fixes and improvements" I can just remove that item.

PeterTurcan:
Yes, I see no value in a remark that vague, so best to remove the line.

PR #621, opened 2026-04-19
Add general note about StaticAssert, notes for Atomic, Filesystem, Log

Lastique:


PeterTurcan (commented):
Good clear release notes.

Can you verify the Filesystem notes are in priority order, in particular the top three are the most important changes?

Lastique:
I don't order release notes by importance. Importance is subjective. The order is the same as in Boost.Filesystem release notes, which is more or less chronologic, although some topic grouping may take place.

PR #622, opened 2026-04-20
Add release notes for: Charconv, Math, Multiprecision, Random

mborland:


PeterTurcan (changes requested):
Fixed in what sense - to come up with character alternatives?

PeterTurcan (changes requested):
Not sure what this means, adding a "finding" parameter or finding parameters of some sort?

Added parameter finding for `non_central_t_distribution`

Are the notes in priority order?  Most impactful three should come first.

mborland:
> Fixed in what sense - to come up with character alternatives?

As in it now works as expected

mborland:
> Not sure what this means, adding a "finding" parameter or finding parameters of some sort?
> 
> Added parameter finding for `non_central_t_distribution`
> 
> Are the notes in priority order?  Most impactful three should come first.

It is used to find the parameters. Yes, I like to do additions first because those are more notable than bug fixes.

PR #630, opened 2026-05-12
Planning Review for Int128 for Matt Borland

Becheler:
Planning Review for Int128 for Matt Borland


mborland:
Capy / Corosio review is now scheduled from 23 June - 7 July. We'll have to push this to later in July

mborland:
@PeterTurcan this should be good to go. I thought it would be improper to add my own review to the schedule.

PR #643, opened 2026-06-29
Update release notes for Boost 1.92.0 Geometry library

vissarion:
Added improvements and solved issues for the Geometry library, including dropping varray implementation and enhancing accuracy.


PeterTurcan (changes requested):
Can you put `glyphs` around keywords so it is clear they are code - varray, static_vector, Andoyer inverse?, index_rtree_exceptions_rst  ?

PR #654, opened 2026-07-13
Update boost_1_92_0.adoc with Conversion, DLL and LexicalCast changes

apolukhin:


PeterTurcan (changes requested):
Should smart_library and shared_library be in glyphs - are they code?

smart_library constructors from shared_library

such as `smart_library` ?

PR #664, opened 2026-07-22
Add @antora/collector-extension to the playbooks

sgerbino:
Registers the official Antora Collector extension so libraries can import files that live outside their doc/ content root (compiled example sources, tested snippet files) into the content catalog via an ext.collector scan declared in their own antora.yml.

Libraries that do not declare collector config are unaffected; the extension is a no-op for them. First consumer is Boost.Capy, whose documentation code blocks are included from sources compiled and run by its test suite.

### DIFF
--- a/libs.playbook.yml
+++ b/libs.playbook.yml
@@ -49,6 +49,7 @@ antora:
         page-boost-branch: develop
         page-boost-ui-branch: develop
         page-commit-id: '000000'
+    - require: '@antora/collector-extension'
     - require: '@cppalliance/antora-cpp-tagfiles-extension'
       cpp-tagfiles:
         using-namespaces:
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "packages": {
     "": {
       "dependencies": {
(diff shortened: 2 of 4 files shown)


sgerbino:
Closing in favor of using symlinks as recommended by @mborland, shown in Boost.Decimal.

alandefreitas:
Nice. Writing a mrdocs extension to build examples from docs is actually easy (I think there's even an example in the mrdocs documentation). You might also be interested in that because many examples actually live there. 

But, more to the point, I also wrote a very similar Antora extension for mrdocs. It's local to mrdocs. I didn't know Antora created a public one since then. Nice. I think it'd be nice to compare our approaches in the PR since we're creating a convention for all boost libraries now. 

Also note that your PR would only affect accepted boost libraries that also share this common playbook (many decided not to). In other words,
this is going to have no effect on Capy. You're probably aware of this and just preparing for the future. But it's worth mentioning it anyway because quite a few people use the UI bundle and think extensions from this common playbook would be inherited by their local Antora system.

PR #668, opened 2026-07-29
Add Boost.Multi Mini-Review to the Review Calendar

mborland:


PeterTurcan:
@mborland - looks like you have an extra "|" between the lib and docs links?

mborland:
@PeterTurcan should be fixed. Something else that I've thought about before, do you think we should remove the results column at the end since it will always be TBD? The format is different between the pending and completed results anyway, so it's not like we are currently copying-pasting and then updating the TBD.

PeterTurcan:
@mborland Seems fair, I will remove the Result column shortly

PR #670, opened 2026-07-30
Current Schedule table of reviews cleaned up

PeterTurcan:
@mborland Cleaned up the Current Review Schedule table - removed the Result, tweaked the column sizes, divided Lib and Docs column into two columns. Looks neater to me but feedback welcome!


mborland (commented):
Definitely looks like an improvement to me. Separating the doc and repo links was a good choice.

PR #680, opened 2026-08-08
added asio release notes

joaquintides:
Important: this is not coming from the author @chriskohlhoff, please have it on hold and merge only at the latest moment if Chris does not propose his own PR.


joaquintides:
#682

